PR #40, opened 2021-02-01
Update API to work with latest transformers/datasets package

SeanNaren:
## What does this PR do?

We shouldn't rely on such an older version for transformers/datasets. This will require re-training the model (which I've done locally, Once approved I can upload the new model weights to S3).

### DIFF
--- a/flash/core/classification.py
+++ b/flash/core/classification.py
@@ -22,11 +22,9 @@ from flash.core.model import Task
 
 class ClassificationDataPipeline(TaskDataPipeline):
 
-    def before_uncollate(self, batch: Union[torch.Tensor, tuple, SequenceClassifierOutput]) -> torch.Tensor:
+    def before_uncollate(self, batch: Union[torch.Tensor, tuple]) -> torch.Tensor:
         if isinstance(batch, tuple):
             batch = batch[0]
-        elif isinstance(batch, SequenceClassifierOutput):
-            batch = batch.logits
         return torch.softmax(batch, -1)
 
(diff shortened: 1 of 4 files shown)


flash/core/classification.py
```diff
@@ -11,19 +11,22 @@
...
 from flash.core.model import Task
 
 
 class ClassificationDataPipeline(TaskDataPipeline):
 
-    def before_uncollate(self, batch: torch.Tensor) -> torch.Tensor:
+    def before_uncollate(self, batch: Union[torch.Tensor, tuple, SequenceClassifierOutput]) -> torch.Tensor:
         if isinstance(batch, tuple):
             batch = batch[0]
+        elif isinstance(batch, SequenceClassifierOutput):
```
carmocca, line 28: Shouldn't this be in `TextClassificationPipeline` instead?

SeanNaren: I didn't even realise that was a thing, yes it should be!

teddykoker (approved):
🔥 🔥

flash/core/classification.py
```diff
@@ -11,17 +11,18 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-from typing import Any
+from typing import Any, Union
 
 import torch
+from transformers.modeling_outputs import SequenceClassifierOutput
```
carmocca, line 17: Unused import

carmocca (approved):

PR #42, opened 2021-02-01
Custom Task Tutorial

teddykoker:
TODO


docs/source/custom_task.rst
```diff
@@ -1,37 +1,139 @@
...
-You can create your own Flash task to if you want to solve any other deep learning problem.
+In this tutorial we will go over the process of creating a custom task,
+along with a custom data module.
 
-See this example for defining a linear classifier task:
-
-.. testcode::
+.. code:: ipython3
 
+    import flash
```
aribornstein, line 9: You should introduce that you are importing the necessary dependencies. Why do we need sklearn?

docs/source/custom_task.rst
```diff
@@ -1,37 +1,139 @@
...
-            super().__init__(model=None,
+    from torch.utils.data import TensorDataset, DataLoader
+    from torch import nn
+    from sklearn import datasets
+    from sklearn.model_selection import train_test_split
+
+The Task
+--------
+
+Here we create a basic linear regression task by subclassing
```
aribornstein, line 20: Put a link to explaining what linear regression is

docs/source/custom_task.rst
```diff
@@ -1,37 +1,139 @@
...
                 metrics=metrics,
                 learning_rate=learning_rate,
             )
-            self.save_hyperparameters()
+            
+        def forward(self, x):
+            return self.model(x)
 
-            self.linear = torch.nn.Linear(num_inputs, num_classes)
+Where is the training step?
```
aribornstein, line 48: Where **are** the training step?

docs/source/custom_task.rst
```diff
@@ -1,37 +1,139 @@
...
+Because of our custom data pipeline’s ``after_uncollate`` method, we
+will get a nicely formatted output like the following:
+
+::
+
+   ['disease progression: 155.90',
+    'disease progression: 156.59',
+    'disease progression: 152.69',
+    'disease progression: 149.05',
+    'disease progression: 150.90']
```
tchaton, line 139: Let's add the DataPipeline object to the doc.

tchaton (approved):
LGTM !

docs/source/custom_task.rst
```diff
@@ -1,37 +1,139 @@
...
 
-            self.linear = torch.nn.Linear(num_inputs, num_classes)
+Where is the training step?
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-        def forward(self, x):
-            return self.linear(x)
+Most models can be trained simply by passing the output of ``forward``
+to the supplied ``loss_fn``, and then passing the resulting loss to the
+supplied ``optimizer``. If you need a more custom configuration, you can
```
aribornstein, line 53: This is not so accessible. Make it one level more accessible

aribornstein (changes requested):
This is pretty clear if you know lightning but I don't think it's accesible to a flash audience. I think many people will have trouble extrapolating this process to their own models  if you want to sit together I can help you make the tone and content more accesible.

docs/source/custom_task.rst
```diff
@@ -1,37 +1,139 @@
...
-            metrics=[Accuracy()],
-            learning_rate=1e-3,
-        ):
-            super().__init__(model=None,
+    from torch.utils.data import TensorDataset, DataLoader
+    from torch import nn
+    from sklearn import datasets
+    from sklearn.model_selection import train_test_split
+
+The Task
```
carmocca, line 17:
```suggestion
The Task: Linear regression
```

docs/source/custom_task.rst
```diff
@@ -1,37 +1,139 @@
...
+--------
+
+Here we create a basic linear regression task by subclassing
+``flash.Task``. For the majority of tasks, you will likely only need to
+override the ``__init__`` and ``__forward__`` methods of task.
```
carmocca, line 22:
```suggestion
override the ``__init__`` and ``__forward__`` methods.
```

docs/source/custom_task.rst
```diff
@@ -1,37 +1,139 @@
...
+                model=model,
                 loss_fn=loss_fn,
                 optimizer=optimizer,
                 metrics=metrics,
                 learning_rate=learning_rate,
             )
-            self.save_hyperparameters()
+            
+        def forward(self, x):
+            return self.model(x)
```
carmocca, line 46:
```suggestion
            # we don't actually need to override this method for this example
            return self.model(x)
```

docs/source/custom_task.rst
carmocca: Where are the training/validation/test steps?

docs/source/custom_task.rst
```diff
@@ -1,37 +1,139 @@
...
+
+For a task you will likely need a specific way of loading data. For this
+example, lets say we want a ``flash.DataModule`` to be used explicitly
+for the prediction of diabetes disease progression. We can create this
+``DataModule`` below, wrapping the scikit-learn `Diabetes
+dataset <https://scikit-learn.org/stable/datasets/toy_dataset.html#diabetes-dataset>`__.
+
+.. code:: ipython3
+
+    class DiabetesPipeline(flash.core.data.TaskDataPipeline):
```
carmocca, line 71: I guess we should introduce the pipeline before mentioning it here

edenlightning:
I suggest something in this pattern: for each object and the API explain what should be added, returned, and why.

The Task
========

To create a custom task you will need:

1. A model- this can be an :class:`~torch.nn.module` or a :class:`~pytorch_lightning.LightningModule`.

2. A loss function

3. An optimizer.

To create the task, subclass the :class:`~flash.core.model.Task`. For the majority of tasks, you will likely only need to
override the :func:`~flash.core.model.Task.__init__` and :func:`~flash.core.model.Task.forward` methods.


* The :func:`~flash.core.model.Task.__init__` method should set the model, the optimizer and the loss function you want to use in your task, and pass them to `super`.

.. code:: ipython3

    class LinearRegression(flash.Task):
        def __init__(self, num_inputs, learning_rate=0.001, metrics=None):

            # what kind of model do we want?
            model = nn.Linear(num_inputs, 1)
    
            # what loss function do we want?
            loss_fn = torch.nn.functional.mse_loss
            
            # what optimizer to do we want?
            optimizer = torch.optim.SGD
            
            super().__init__(
                model=model,
                loss_fn=loss_fn,
                optimizer=optimizer,
                metrics=metrics,
                learning_rate=learning_rate,
            )

* The step method.....


.. code:: ipython3


* The :func:`~flash.core.model.Task.forward` method will define the operations you want to use for prediction.

.. code:: ipython3

    #TODO: add a better example here
....


* The predict method ...


* Data pipeline...

PR #69, opened 2021-02-04
Docs fix

edenlightning:
Fix various parts of docs


discdiver:
This looks great! With these changes it would be clear  the text summarization won't run on a CPU - which would have saved me time. (Although the docs say Flash should work on CPU, GPU, and TPU, so maybe  a new issue is warranted 😉).

Borda (approved):

PR #109, opened 2021-02-12
fix: set inputs as optional

### LINKED ISSUE #108: Make categorical_input/numerical_input optional for TabularData.from_df
## 🐛 Bug
If I have only numerical features then I still have to pass an empty list for the categorical_input. It can be optional.
image

### To Reproduce

1. df = pd.DataFrame({'digit': [1,2,3], 'odd_even':[0,1,0]})
2. datamodule = TabularData.from_df(df, 'odd_even', 
                    numerical_input=['digit'],
                   )

```
TypeError                                 Traceback (most recent call last)
<ipython-input-122-405a8bb49976> in <module>
      1 datamodule = TabularData.from_df(final_data, 'target', 
----> 2                     numerical_input=train_x.columns.tolist(),
      3 #                     categorical_input=[],
      4                    )

TypeError: from_df() missing 1 required positional argument: 'categorical_input'
```

#### Code sample
```
df = pd.DataFrame({'digit': [1,2,3], 'odd_even':[0,1,0]})
datamodule = TabularData.from_df(df, 'odd_even', 
                    numerical_input=['digit'],
                   )
```

### Expected behaviour
If only one of categorical or numerical input is passed then users should not be forced to enter an empty list.

### Environment

 - PyTorch Version (e.g., 1.0): '1.7.1'
 - OS (e.g., Linux): MacOS
 - How you installed PyTorch (`conda`, `pip`, source): pip
 - Python version:Python 3.7.9

> I would love to start my contribution to Flash by fixing this issue.

aniketmaurya:
## What does this PR do?

`categorical_inputs` and `numerical_inputs` are made optional. The error will raise only when both are None.
Fixes #108 

## Before submitting

## PR review

## Did you have fun?


flash/tabular/classification/data/data.py
```diff
@@ -82,6 +82,15 @@ def __init__(
         dfs = [train_df]
         self._test_df = None
 
+        if categorical_input is None and numerical_input is None:
+            raise TypeError('Both categorical_input and numerical_input are None!')
```
kaushikb11, line 86: It's not `TypeError`, as the argument received is of the appropriate type. You could have `from pytorch_lightning.utilities.exceptions import MisconfigurationException` or `ValueError` instead.

flash/tabular/classification/data/data.py
```diff
@@ -82,6 +82,15 @@ def __init__(
...
+
+        if categorical_input is None:
+            categorical_input = []
+
+        if numerical_input is None:
```
kaushikb11, line 91: Why not just have `[]` as default inputs?

kaushikb11:
@aniketmaurya Could you add a test as well for this edge case? It should be able to raise an Exception.

flash/tabular/classification/data/data.py
aniketmaurya: Yes, you are right @kaushikb11 . one more thing, instead of List shouldn't we have Tuple? Since these values are not going to modify.

aniketmaurya:
> @aniketmaurya Could you add a test as well for this edge case? It should be able to raise an Exception.

Ok will do

flash/tabular/classification/data/data.py
Borda, line 86:
```suggestion
            raise RuntimeError('Both `categorical_input` and `numerical_input` are None!')
```

Borda (approved):

teddykoker:
Usually it's frowned upon to have empty lists/tuples as default arguments, see: https://nikos7am.com/posts/mutable-default-arguments/

I would make the default values `None` and pass an empty list if the value is None at runtime instead

flash/tabular/classification/data/data.py
```diff
@@ -194,8 +203,8 @@ def from_csv(
         cls,
         train_csv: str,
         target: str,
-        categorical_input: List,
-        numerical_input: List,
+        categorical_input: List = (),
```
teddykoker, line 206: See comment above about mutable defaults

teddykoker (changes requested):

flash/tabular/classification/data/data.py
aniketmaurya: I may sound dumb but `()` is a tuple and immutable, right?

aniketmaurya: thanks for sharing the article link. I have done the suggested change

teddykoker (approved):
Awesome thanks :)

akihironitta (approved):

kaushikb11 (approved):

PR #130, opened 2021-02-18
Datapipeline poc

### LINKED ISSUE #67: Improve DataPipeline API
### Motivation

* Should we DataPipeline by splitted in 2, collate_fn and uncollate_fn ?

Reason:
- collate_fn is really coming from the dataset to process new raw_data
- uncollate_fn is usually related to a task such as classification. It can be confusing.

default DataPipeline

```
class DataPipeline(CollatePipeline, UnCollatePipeline):

    ...

 
```

create text classification data pipeline.

```
class TextClassificationDataPipeline(TextCollatePipeline, ClassificationUnCollatePipeline):

    ...

 
```

* Create BaseDataModule per data-type such as TextDataModule, ImageDataModule + their associated TextDataPipeline, ImageDataPipeline as default.

justusschock:
## What does this PR do?
This is just some API prototype. So far it is not completely working. It is basically just meant as a discussion starter :)

Fixes #67 

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/flash/core/model.py
+++ b/flash/core/model.py
@@ -17,10 +17,13 @@ from typing import Any, Callable, Dict, Mapping, Optional, Sequence, Type, Union
 
 import pytorch_lightning as pl
 import torch
+from pytorch_lightning import Trainer
 from torch import nn
 
-from flash.core.data import DataModule, DataPipeline
+from flash.core.data import DataModule
 from flash.core.utils import get_callable_dict
+from flash.data.data_pipeline import DataPipeline
+from flash.data.postprocessing_pipeline import PostProcessingPipeline
 
 
 def predict_context(func: Callable) -> Callable:
@@ -31,13 +34,16 @@ def predict_context(func: Callable) -> Callable:
 
     @functools.wraps(func)
     def wrapper(self, *args, **kwargs) -> Any:
+        grad_enabled = torch.is_grad_enabled()
+        is_training = self.training
         self.eval()
         torch.set_grad_enabled(False)
 
         result = func(self, *args, **kwargs)
 
-        self.train()
-        torch.set_grad_enabled(True)
+        if is_training:
+            self.train()
+        torch.set_grad_enabled(grad_enabled)
         return result
 
     return wrapper
@@ -63,6 +69,8 @@ class Task(pl.LightningModule):
         learning_rate: float = 5e-5,
     ):
         super().__init__()
+        self._last_trainer_kwargs = {}
+
         if model is not None:
             self.model = model
         self.loss_fn = {} if loss_fn is None else get_callable_dict(loss_fn)
@@ -144,7 +152,7 @@ class Task(pl.LightningModule):
 
         """
         # enable x to be a path to a folder
-        if isinstance(x, str):
+        if isinstance(x, str) and os.path.isdir(x):
             files = os.listdir(x)
             files = [os.path.join(x, y) for y in files]
             x = files
@@ -163,22 +171,36 @@ class Task(pl.LightningModule):
     def data_pipeline(self) -> DataPipeline:
         # we need to save the pipeline in case this class
         # is loaded from checkpoint and used to predict
-        if not self._data_pipeline:
-            try:
-                # datamodule pipeline takes priority
-                self._data_pipeline = self.trainer.datamodule.data_pipeline
-            except AttributeError:
-                self._data_pipeline = self.default_pipeline()
-        return self._data_pipeline
+        return self._get_pipeline('data')
 
     @data_pipeline.setter
     def data_pipeline(self, data_pipeline: DataPipeline) -> None:
         self._data_pipeline = data_pipeline
 
+    @property
+    def postprocessing_pipeline(self) -> PostProcessingPipeline:
+        return self._get_pipeline('postprocessing')
+
+    def _get_pipeline(self, pipeline_type: str):
+        pipeline_attr_name = f'{pipeline_type}_pipline'
+
+        if getattr(self, '_' + pipeline_attr_name) is not None:
+            return getattr(self, '_' + pipeline_attr_name)
+
+        if self.datamodule is not None and hasattr(self, pipeline_attr_name):
+            return getattr(self.datamodule, pipeline_attr_name)
+
+        if self.trainer is not None and hasattr(self.trainer, 'datamodule') and self.trainer.datamodule is not None:
+            if hasattr(self.trainer.datamodule,
+                       pipeline_attr_name) and getattr(self.trainer.datamodule, pipeline_attr_name is not None):
+                return getattr(self.trainer.datamodule, pipeline_attr_name is not None)
+
+        return None
+
     @staticmethod
-    def default_pipeline() -> DataPipeline:
+    def default_data_pipeline() -> DataPipeline:
         """Pipeline to use when there is no datamodule or it has not defined its pipeline"""
-        return DataModule.default_pipeline()
+        return DataModule.default_data_pipeline()
 
     def on_load_checkpoint(self, checkpoint: Dict[str, Any]) -> None:
         self.data_pipeline = checkpoint["pipeline"]
@@ -188,3 +210,111 @@ class Task(pl.LightningModule):
 
     def configure_finetune_callback(self):
         return []
+
+    ### THE FOLLOWING IS A POC FOR DISTRIBUTED PREDICTION
+    def on_predict_start(self):
+        # TODO: Add hook to lightning Trainer
+        if self.data_pipeline is not None:
+            self.data_pipeline._attach_to_model(self)
+
+        if self.postprocessing_pipeline is not None:
+            self.postprocessing_pipeline._attach_to_model(self)
+
+    def predict_step(self, batch, batch_idx):
+        # TODO: Move lightning predict loop from predict to predict_step
+        if isinstance(batch, (tuple, list)) and len(batch) == 2:
+            x, y = batch
+        else:
+            x, y = batch, None
+
+        return self(x)
+
+    def new_predict(
+        self,
+        x: Any,
+        skip_collate: Optional[bool] = None,
+        data_pipeline: Optional[DataPipeline] = None,
+        postprocessing_pipeline: Optional[PostProcessingPipeline] = None,
+        data_loader_kwargs: Optional[dict] = None,
+        **trainer_kwargs
+    ):
+        if data_pipeline is not None:
+            self.data_pipeline = data_pipeline
+        if postprocessing_pipeline is not None:
+            self.postprocessing_pipeline = postprocessing_pipeline
+
+        trainer = self._create_trainer('predict', **trainer_kwargs)
+
+        if data_loader_kwargs is None:
+            data_loader_kwargs = {}
+
+        if 'num_workers' not in data_loader_kwargs:
+            # leave one for main process
+            data_loader_kwargs['num_workers'] = os.cpu_count() - 1
+
+        auto_collate = None
+        if 'collate_fn' not in data_loader_kwargs:
+            auto_collate = not skip_collate
+
+        dl = self.data_pipeline._generate_loader(x, auto_collate=auto_collate, **data_loader_kwargs)
+
+        return trainer.predict(self, dl)
+
+    def _create_trainer(self, stage: str, **trainer_kwargs):
+        # TODO: Also use these for trainer creation in training?
+        # TODO: Have default trainer kwargs per task?
+        _trainer_kwargs = {}
+        # TODO: Adjust this to trainer running stage from pl
+        if stage == 'predict':
+            _trainer_kwargs.update(logger=None)
+
+        if not 'gpus' in trainer_kwargs and not 'tpu_cores' in trainer_kwargs:
+            _trainer_kwargs['gpus'], _trainer_kwargs['tpu_cores'] = self._parse_default_devices()
+
+        _trainer_kwargs.update(trainer_kwargs)
+
+        if not hasattr(self, 'trainer') or self.trainer is None or self._last_trainer_kwargs != trainer_kwargs:
+            self._last_trainer_kwargs = _trainer_kwargs
+            self.trainer = None
+            return Trainer(**_trainer_kwargs)
+
+        else:
+            return self.trainer
+
+    def _parse_default_devices(self):
+        gpus = None,
+        tpu_cores = None
+
+        if torch.cuda.is_available():
+            gpus = torch.cuda.device_count()
+
+        # TODO: Add logic for automatted TPU device parsing
+
+        return gpus, tpu_cores
+
+    def serve(
+        self,
+        x,
+        skip_collate: Optional[bool] = None,
+        data_pipeline: Optional[DataPipeline] = None,
+        postprocessing_pipeline: Optional[PostProcessingPipeline] = None,
+        data_loader_kwargs: Optional[dict] = None,
+        **trainer_kwargs
+    ):
+        """Serving for Production. Basically same as prediction, just other defaults (no workers, no distributed prediction)
+        """
+
+        if data_loader_kwargs is None:
+            data_loader_kwargs = {}
+        data_loader_kwargs['num_workers'] = 0
+
+        trainer_kwargs['num_gpus'] = [0] if torch.cuda.is_available() else 0
+        # TODO: tpu_cores
+        return self.new_predict(
+            x,
+            skip_collate=skip_collate,
+            data_pipeline=data_pipeline,
+            postprocessing_pipeline=postprocessing_pipeline,
+            data_loader_kwargs=data_loader_kwargs,
+            **trainer_kwargs
+        )
--- a/flash/data/data_pipeline.py
+++ b/flash/data/data_pipeline.py
@@ -1,11 +1,11 @@
-from typing import Any, Callable, Iterable, Optional, Sequence, Tuple, Union
-import torch
 from functools import wraps
-from torch.utils.data.dataloader import default_collate, DataLoader
-from pytorch_lightning.core import LightningModule
+from typing import Any, Callable, Iterable, Optional, Sequence, Tuple, Union
 
-from pytorch_lightning.utilities.exceptions import MisconfigurationException
+import torch
+from pytorch_lightning.core import LightningModule
 from pytorch_lightning.trainer.connectors.data_connector import _PatchDataLoader
+from pytorch_lightning.utilities.exceptions import MisconfigurationException
+from torch.utils.data.dataloader import DataLoader, default_collate
 
 
 class DataPipeline:
@@ -22,7 +22,7 @@ class DataPipeline:
 
     def post_collate(self, batch: Any) -> Any:
         """Transforms to apply to a whole batch (if possible use this for efficiency)
-        
+
         .. note::
             This option is mutually exclusive with :meth:`device_pre_collate`, since if both are specified, uncollation has to be applied.
         """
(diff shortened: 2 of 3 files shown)


flash/core/model.py
```diff
@@ -163,22 +171,36 @@ def configure_optimizers(self) -> torch.optim.Optimizer:
...
+        return self._get_pipeline('data')
 
     @data_pipeline.setter
     def data_pipeline(self, data_pipeline: DataPipeline) -> None:
         self._data_pipeline = data_pipeline
 
+    @property
+    def postprocessing_pipeline(self) -> PostProcessingPipeline:
+        return self._get_pipeline('postprocessing')
+
```
justusschock, line 183: TODO: Missing setter

flash/core/model.py
```diff
@@ -163,22 +171,36 @@ def configure_optimizers(self) -> torch.optim.Optimizer:
...
+                return getattr(self.trainer.datamodule, pipeline_attr_name is not None)
+
+        return None
+
     @staticmethod
-    def default_pipeline() -> DataPipeline:
+    def default_data_pipeline() -> DataPipeline:
         """Pipeline to use when there is no datamodule or it has not defined its pipeline"""
-        return DataModule.default_pipeline()
+        return DataModule.default_data_pipeline()
```
justusschock, line 203: TODO: also do this for postprocessing

flash/core/model.py
```diff
@@ -188,3 +210,111 @@ def on_save_checkpoint(self, checkpoint: Dict[str, Any]) -> None:
 
     def configure_finetune_callback(self):
         return []
+
+    ### THE FOLLOWING IS A POC FOR DISTRIBUTED PREDICTION
+    def on_predict_start(self):
```
justusschock, line 215: @tchaton  does it make sense to have a hook like that (I think we need to revisit lightning hooks in general for all stages)

flash/core/model.py
```diff
@@ -188,3 +210,111 @@ def on_save_checkpoint(self, checkpoint: Dict[str, Any]) -> None:
...
+        # TODO: Add hook to lightning Trainer
+        if self.data_pipeline is not None:
+            self.data_pipeline._attach_to_model(self)
+
+        if self.postprocessing_pipeline is not None:
+            self.postprocessing_pipeline._attach_to_model(self)
+
+    def predict_step(self, batch, batch_idx):
+        # TODO: Move lightning predict loop from predict to predict_step
```
justusschock, line 224: @tchaton You mentioned the prediction API is not final in lightning, right?

IMO it makes sense to rename it to training_step within the `LightningModule`, since (similar to train step etc.) it only runs prediction for one batch at a time, making it more of a step (plus we can use the `predict` keyword here independently :) )

flash/core/model.py
```diff
@@ -188,3 +210,111 @@ def on_save_checkpoint(self, checkpoint: Dict[str, Any]) -> None:
...
+
+        dl = self.data_pipeline._generate_loader(x, auto_collate=auto_collate, **data_loader_kwargs)
+
+        return trainer.predict(self, dl)
+
+    def _create_trainer(self, stage: str, **trainer_kwargs):
+        # TODO: Also use these for trainer creation in training?
+        # TODO: Have default trainer kwargs per task?
+        _trainer_kwargs = {}
+        # TODO: Adjust this to trainer running stage from pl
```
justusschock, line 267: any thoughts on that @tchaton  @aribornstein ?

flash/data/data_pipeline.py
```diff
@@ -0,0 +1,224 @@
...
+        if isinstance(data, Iterable) and self.is_overriden('load_sample'):
+            load_per_sample = True
+            load_fn = self.load_sample
+        else:
+            load_per_sample = False
+            load_fn = self.load_data
+
+        return AutoDataset(data=data, load_fn=load_fn, load_per_sample=load_per_sample)
+
+    def _generate_loader(
```
justusschock, line 170: should me make this public API? @tchaton

flash/data/data_pipeline.py
```diff
@@ -0,0 +1,224 @@
...
+                    dataloader = _PatchDataLoader(dataloader)
+
+                setattr(model, loader_name, dataloader)
+
+        model.transfer_batch_to_device = (
+            self._model_transfer_to_device_wrapper(model.transfer_batch_to_device, device_collater)
+        )
+        return model
+
+    def _generate_auto_dset(self, data: Union[Iterable, Any]) -> AutoDataset:
```
justusschock, line 160: should me make this public API? @tchaton

flash/core/model.py
```diff
@@ -163,22 +171,36 @@ def configure_optimizers(self) -> torch.optim.Optimizer:
...
     @data_pipeline.setter
     def data_pipeline(self, data_pipeline: DataPipeline) -> None:
         self._data_pipeline = data_pipeline
 
+    @property
+    def postprocessing_pipeline(self) -> PostProcessingPipeline:
+        return self._get_pipeline('postprocessing')
+
+    def _get_pipeline(self, pipeline_type: str):
+        pipeline_attr_name = f'{pipeline_type}_pipline'
```
tchaton, line 185: `_pipline ` typo ?

flash/core/model.py
```diff
@@ -163,22 +171,36 @@ def configure_optimizers(self) -> torch.optim.Optimizer:
...
+        if self.trainer is not None and hasattr(self.trainer, 'datamodule') and self.trainer.datamodule is not None:
+            if hasattr(self.trainer.datamodule,
+                       pipeline_attr_name) and getattr(self.trainer.datamodule, pipeline_attr_name is not None):
+                return getattr(self.trainer.datamodule, pipeline_attr_name is not None)
+
+        return None
+
     @staticmethod
-    def default_pipeline() -> DataPipeline:
+    def default_data_pipeline() -> DataPipeline:
```
tchaton, line 201: I think here we should take the data-type default one ? Example collate for text isn't the same than for vision.

flash/core/model.py
```diff
@@ -163,22 +171,36 @@ def configure_optimizers(self) -> torch.optim.Optimizer:
...
+
+        if getattr(self, '_' + pipeline_attr_name) is not None:
+            return getattr(self, '_' + pipeline_attr_name)
+
+        if self.datamodule is not None and hasattr(self, pipeline_attr_name):
+            return getattr(self.datamodule, pipeline_attr_name)
+
+        if self.trainer is not None and hasattr(self.trainer, 'datamodule') and self.trainer.datamodule is not None:
+            if hasattr(self.trainer.datamodule,
+                       pipeline_attr_name) and getattr(self.trainer.datamodule, pipeline_attr_name is not None):
```
tchaton, line 195: When can `pipeline_attr_name ` be None ?

flash/core/model.py
tchaton: Where is it called ? I guess we could add hook for predict. Need a bit more exploration there.

flash/core/model.py
tchaton: Yes, I am good with that :)

flash/core/model.py
```diff
@@ -188,3 +210,111 @@ def on_save_checkpoint(self, checkpoint: Dict[str, Any]) -> None:
...
+        if self.data_pipeline is not None:
+            self.data_pipeline._attach_to_model(self)
+
+        if self.postprocessing_pipeline is not None:
+            self.postprocessing_pipeline._attach_to_model(self)
+
+    def predict_step(self, batch, batch_idx):
+        # TODO: Move lightning predict loop from predict to predict_step
+        if isinstance(batch, (tuple, list)) and len(batch) == 2:
+            x, y = batch
```
tchaton, line 226: You shouldn't have y when doing predictions.

flash/core/model.py
tchaton: I like it. We had similar function in previous iteration of predict.

flash/core/model.py
```diff
@@ -188,3 +210,111 @@ def on_save_checkpoint(self, checkpoint: Dict[str, Any]) -> None:
...
+        tpu_cores = None
+
+        if torch.cuda.is_available():
+            gpus = torch.cuda.device_count()
+
+        # TODO: Add logic for automatted TPU device parsing
+
+        return gpus, tpu_cores
+
+    def serve(
```
tchaton, line 295: I think this serve function is confusing. Serve means different things on the context and should not be used to perform prediction. It is more like a mode.

flash/data/data_pipeline.py
tchaton: Yes, and it should be `to_dataloader`.

flash/data/data_pipeline.py
```diff
@@ -0,0 +1,224 @@
...
+
+    def __init__(self, data: Union[Iterable, Any], load_fn: Callable, load_per_sample: bool) -> None:
+        super().__init__()
+
+        self.data = data
+        self.load_fn = load_fn
+
+        self._load_lazy = load_per_sample
+
+        if not self._load_lazy:
```
tchaton, line 214: I think it should be `in_memory`.

flash/data/data_pipeline.py
```diff
@@ -0,0 +1,224 @@
...
+from pytorch_lightning.utilities.exceptions import MisconfigurationException
+from torch.utils.data._utils.collate import default_collate, default_convert
+from torch.utils.data.dataloader import DataLoader
+
+
+class DataPipeline:
+
+    def load_data(self, data: Any) -> Any:
+        """Loads entire data from Dataset"""
+        return data
```
tchaton, line 16:
```suggestion
    def load_data(self, data: Any) -> Any:
        """Loads entire data in-memory from Dataset"""
        return data
```

flash/data/data_pipeline.py
```diff
@@ -0,0 +1,224 @@
...
+    def post_collate(self, batch: Any) -> Any:
+        """Transforms to apply to a whole batch (if possible use this for efficiency)
+
+        .. note::
+            This option is mutually exclusive with :meth:`device_pre_collate`, since if both are specified, uncollation has to be applied.
+        """
+        return batch
+
+    def device_pre_collate(self, sample: Any) -> Any:
+        """Transforms to apply to the data before the collation (per-sample basis).
```
tchaton, line 35:
```suggestion
    def on_device_pre_collate(self, sample: Any) -> Any:
        """Transforms to apply to the data before the collation (per-sample basis).
```

flash/data/data_pipeline.py
```diff
@@ -0,0 +1,224 @@
...
+        .. note::
+            This option is mutually exclusive with :meth:`post_collate`, since if both are specified, uncollation has to be applied.
+
+        .. note::
+            This function won't be called within the dataloader workers, since to make that happen each of the workers would have to create it's own CUDA-context which would pollute GPU memory (if on GPU).
+        """
+        return sample
+
+    def device_post_collate(self, batch: Any) -> Any:
+        """
```
tchaton, line 46:
```suggestion
    def on_device_post_collate(self, batch: Any) -> Any:
        """
```

flash/data/data_pipeline.py
tchaton: I think `to_dataloader ` is enough.

flash/data/data_pipeline.py
```diff
@@ -0,0 +1,224 @@
...
+            worker_collate = collate_fn
+            device_collate = self._do_nothing_collate
+
+        elif device_pre_collate_overriden:
+            worker_collate = self._do_nothing_collate
+            device_collate = collate_fn
+
+        else:
+            worker_collate = collate_fn
+            device_collate = self._do_nothing_collate
```
tchaton, line 92:
```suggestion
        if device_pre_collate_overriden:
            worker_collate = self._do_nothing_collate
            device_collate = collate_fn

        else:
            worker_collate = collate_fn
            device_collate = self._do_nothing_collate
```

flash/data/data_pipeline.py
```diff
@@ -0,0 +1,224 @@
...
+                    wrap_patch_loader = False
+
+                if isinstance(dataloader, Sequence):
+                    was_seq = True
+                else:
+                    dataloader = [dataloader]
+                    was_seq = False
+
+                for idx, loader in enumerate(dataloader):
+                    if isinstance(loader, DataLoader):
```
tchaton, line 136: This won't work with custom dataloader. See data_loading.py in Lightning.

flash/data/data_pipeline.py
```diff
@@ -0,0 +1,224 @@
...
+
+                if not was_seq:
+                    dataloader = dataloader[0]
+
+                if wrap_patch_loader:
+                    dataloader = _PatchDataLoader(dataloader)
+
+                setattr(model, loader_name, dataloader)
+
+        model.transfer_batch_to_device = (
```
tchaton, line 155: Love this ! Pretty smart !

flash/data/data_pipeline.py
```diff
@@ -0,0 +1,224 @@
...
+                raise MisconfigurationException('auto_collate and collate_fn are mutually exclusive')
+
+        else:
+            if auto_collate is None:
+                auto_collate = True
+
+            if auto_collate:
+                loader_kwargs['collate_fn'] = default_collate
+            else:
+                loader_kwargs['collate_fn'] = default_convert
```
tchaton, line 184: Isn't `default_convert` used only for numpy array ?

flash/core/model.py
```diff
@@ -163,22 +171,36 @@ def configure_optimizers(self) -> torch.optim.Optimizer:
...
-            try:
-                # datamodule pipeline takes priority
-                self._data_pipeline = self.trainer.datamodule.data_pipeline
-            except AttributeError:
-                self._data_pipeline = self.default_pipeline()
-        return self._data_pipeline
+        return self._get_pipeline('data')
 
     @data_pipeline.setter
     def data_pipeline(self, data_pipeline: DataPipeline) -> None:
```
tchaton, line 177: I find it a bit confusing to have `DataPipeline` and `PostProcessingPipeline` as people might expect a `PreprocessingPipeline`. Worth to iterate on this one.

tchaton (commented):
Looks great overall. I would have to covert some current pipeline to this new API and see how it feels.

flash/core/model.py
justusschock: Yeah, but that's why each task would have its own default

flash/core/model.py
justusschock: It can't, that should be outside the brackets :)

flash/core/model.py
justusschock: Similar to on_fit_start, would be called immediately after the Trainer.predict was called

flash/core/model.py
justusschock: Sometimes you have when using the same loader/dart as for training. This is just because it was already there in old predict logic and I didn't want to remove already existing features

flash/core/model.py
justusschock: I k is we had something like that for training in the beginning. The only downside I See, is That it hides away the Lightning Trainer

flash/data/data_pipeline.py
justusschock: That's why it's guarded like this. But IMO we shouldn't expect any custom loaders here, since then people would be using lightning. Also you cannot pitch this for custom loaders without knowing their internals

flash/data/data_pipeline.py
justusschock: No, default convert also converts numbers etc. basically does the same as default_collate without tensor stacking

flash/core/model.py
justusschock: Yes, thought so as well. Basically I named it data_pipeline since it does loading + preprocessing. But fine with changing it as well

flash/core/model.py
aribornstein: Maybe we could provide an optional argument for the user to provide trainer in case they don't want to use the default trainer?

t-vi:
@tchaton pointed me here. I must admit I don't feel that I figure out the design behind the code.

If I may make two observations despite my limited understanding, there are two things where I have the impression that the API you are creating here is not aligned with how I think about my data and my models and how they meet:
- I don't think all preprocessing should be tied to the dataloader (/collate fn). To me "same datamodule, different augmentation" or "I have an image from somewhere (e.g. the webcam) but want some part of preprocessing" can happen and it seems unnatural to have to modify the datamodule for this or having to apply parts manually.
- My impression is that the model here is to deal with stuff passed in to new_predict etc. is to set self.something and then use self as state. As a user this isn't what I'd expect (I would expect that the state is fixed and that I might override parts of it through the args for this one call.).

Part of this might not be solvable within flash itself but might need amending lightning (in particular, there seems no "auxiliary information" going into the train loop/step except the dataloader).

flash/data/data_pipeline.py
aribornstein: I don't agree with that I just ran into a huge issue with this today if I want to extend flash capabilities I shouldn't have to implement a lightning datamodule from scratch to take advantage of lightnings features.

flash/core/model.py
aribornstein: Yes I thought we wanted to reserve serve for actually setting up a hosting server for the model.

justusschock:
@t-vi Thanks for your comments, they are very valuable.

Regarding your first point: 
* Why not tie everything to the loader? Creating a loader (especially one with `num_workers=0`) is almost no overhead.
* In terms of augmentations: You can for example have them as an init argument for your pipeline.
* To the datamodule part: You're right and the integration with the datamodule is definitely not perfect. This should just be a base to iterate on :)

Regarding your second point:
You're right to some extend (And the part where you may not be right, is not visible here so let me explain my thoughts behind that). Yes I kind of wanted to use the model as a state, but only temporary. The part that's still missing here (but should definitely come) is to revert back to the original state.

What kind of API would you expect as a user? Maybe we can look on how we can integrate this kind of API into flash/lightning :)

flash/core/model.py
justusschock: you mean another trainer class?
Actually, the trainer class is something I'd hardcode here tbh.
This is one of the very fundamental lightning aspects and I feel if a user wants to change it, he either should look into customization with callbacks/plugins or subclass the task to overwrite it here directly.

flash/core/model.py
justusschock: okay, so how do we want to handle it? I mean technically the user can set those defaults manually as well, but I feel like this is not what flash should be aiming for.

flash/data/data_pipeline.py
justusschock: Yes, but what kind of interface do you want to assume in that case? E.g. if you have a custom loader class, there might not even be something we can attach to...

flash/core/model.py
carmocca, line 195: `is not None` should be outside of the parenthesis

flash/core/model.py
carmocca: It also does postprocessing, with `after_uncollate`

flash/core/model.py
carmocca: That was my initial proposal but it got turned down because users are expected to write

`model.predict(...)`

but not

`model.predict_step()`

because nobody calls

`model.training_step`

flash/core/model.py
carmocca: Yeah, this should be here in the future

https://pytorch-lightning.readthedocs.io/en/latest/common/lightning_module.html#transfer-batch-to-device

carmocca (commented):
I must be missing lots of context here because this goes in a complete different direction to what we initially discussed for Flash, right?

We agreed on using `model.predict` for simple inference an `trainer.predict` for distributed inference? Has there been further developments about this? What's the user facing API with these changes?

Also what is PostprocessingPipeline? Why is it separate from DataPipeline?

PR #156, opened 2021-03-04
Added predict override for multilabel classification

aribornstein:
## What does this PR do?

Current Multilabel classification PR only has hardcoded categorical predict this PR is attempt to address that limitation


aribornstein:
@justusschock can you merge this into your branch I don't have permissions.

PR #171, opened 2021-03-14
ic backbone nn.Module

williamFalcon:


aribornstein:
This functionality is key but should be applied across all backbones not just at the image classification model layer. This is why I think we need a unified backbone API for tasks.

kaushikb11:
#BYOB Bring your own Backbone! :D

PR #209, opened 2021-04-09
[Typing] add typing for core package 1/n

justusschock:
## What does this PR do?
Trying to add MyPy compatible typing to the core package. There are some issues left, where I didn't know how to resolve them. They can be found by running `mypy -p flash`

@PyTorchLightning/core-flash any idea on how to fix these?

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/flash/core/finetuning.py
+++ b/flash/core/finetuning.py
@@ -11,13 +11,16 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-from typing import List, Union
+from typing import List, TYPE_CHECKING, Union
 
 from pytorch_lightning import LightningModule
 from pytorch_lightning.callbacks import BaseFinetuning
 from pytorch_lightning.utilities.exceptions import MisconfigurationException
 from torch import nn
 from torch.optim import Optimizer
+
+if TYPE_CHECKING:
+    from flash.core.model import Task
 
 
 class NoFreeze(BaseFinetuning):
@@ -37,7 +40,7 @@ class NoFreeze(BaseFinetuning):
 
 class FlashBaseFinetuning(BaseFinetuning):
 
-    def __init__(self, attr_names: Union[str, List[str]] = "backbone", train_bn: bool = True):
+    def __init__(self, attr_names: Union[str, List[str]] = "backbone", train_bn: bool = True) -> None:
         r"""
 
         FlashBaseFinetuning can be used to create a custom Flash Finetuning Callback.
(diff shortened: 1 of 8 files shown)


tchaton (approved):
LGTM !

flash/core/finetuning.py
```diff
@@ -54,17 +57,17 @@ def __init__(self, attr_names: Union[str, List[str]] = "backbone", train_bn: boo
         self.attr_names = [attr_names] if isinstance(attr_names, str) else attr_names
         self.train_bn = train_bn
 
-    def freeze_before_training(self, pl_module: LightningModule) -> None:
+    def freeze_before_training(self, pl_module: 'Task') -> None:
         self.freeze_using_attr_names(pl_module, self.attr_names, train_bn=self.train_bn)
 
-    def freeze_using_attr_names(self, pl_module, attr_names: List[str], train_bn: bool = True):
+    def freeze_using_attr_names(self, pl_module: Task, attr_names: List[str], train_bn: bool = True) -> None:
```
carmocca, line 63:
```suggestion
    def freeze_using_attr_names(self, pl_module: 'Task', attr_names: List[str], train_bn: bool = True) -> None:
```

flash/core/model.py
```diff
@@ -108,20 +108,22 @@ def step(self, batch: Any, batch_idx: int) -> Any:
...
 
-    def forward(self, x: Any) -> Any:
-        return self.model(x)
+    def forward(self, *args: Any, **kwargs: Any) -> Any:
+        return self.model(*args, **kwargs)
 
-    def training_step(self, batch: Any, batch_idx: int) -> Any:
-        output = self.step(batch, batch_idx)
+    # Mypy requires the steps to have this signature since the base class has it and a more explicit
+    # phrasing of the arguments would violate the Liskov substitution principle
```
carmocca, line 115: I prefer if we keep the explicit arguments, and adding `noqa`  if necessary. We want users subclassing this to not have to google the actual arguments they will receive etc.

If this was user code, It would be okay but it's not the case.

flash/core/model.py
```diff
@@ -143,9 +145,16 @@ def predict(
         """
         running_stage = RunningStage.PREDICTING
         data_pipeline = data_pipeline or self.data_pipeline
+
+        # This check is necessary for mypy compatibility
```
carmocca, line 149: does it not work with

`data_pipeline: DataPipeline = data_pipeline or self.data_pipeline`

above?

flash/core/model.py
```diff
@@ -143,9 +145,16 @@ def predict(
...
+        if not isinstance(data_pipeline, DataPipeline):
+            raise TypeError(
+                f'data_pipeline should be of type DataPipeline, but is of type {type(data_pipeline).__name__}!'
+            )
+
         x = [x for x in data_pipeline._generate_auto_dataset(x, running_stage)]
         x = data_pipeline.worker_preprocessor(running_stage)(x)
-        x = self.transfer_batch_to_device(x, self.device)
+        x = self.transfer_batch_to_device(x, torch.device(self.device))
```
carmocca, line 157: I guess the best fix for this would be to add the correct type to Lightning.

Should we create a .pyi stub file instead for these?

flash/core/model.py
justusschock: No, because then it complains that it could still be None

flash/core/model.py
justusschock: We could, I tried to solve it in flash only for now, to avoid touching lightning just for this.

flash/core/model.py
justusschock: Fine with this as well. But we will have many of these then :)

flash/core/model.py
carmocca: the .pyi stub file would live in Flash, until the typing is fixed in Lightning

lightning.pyi

```python
class LightningModule:
    device: torch.Device
```

Not an expert on this, not sure if this works. Copying examples from https://github.com/python/typeshed

flash/core/model.py
carmocca: I know... but that's the case too in Lightning. Not sure if there is a better solution

flash/core/model.py
edgarriba: you can use typing `cast` here

flash/core/finetuning.py
```diff
@@ -54,17 +57,17 @@ def __init__(self, attr_names: Union[str, List[str]] = "backbone", train_bn: boo
         self.attr_names = [attr_names] if isinstance(attr_names, str) else attr_names
         self.train_bn = train_bn
 
-    def freeze_before_training(self, pl_module: LightningModule) -> None:
+    def freeze_before_training(self, pl_module: 'Task') -> None:
```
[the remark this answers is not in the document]
justusschock: the one below is wrong then, thanks :)

flash/core/model.py
justusschock: That's nice. didn't know one could have stub files for other libs included

PR #239, opened 2021-04-22
Add semantic segmentation task

edgarriba:
## What does this PR do?

This PR implements semantic segmentation task. It includes the following data structures:
- `SemantincSegmentationPreprocess`
- `SemantincSegmentationData`
- `SemantincSegmentation`

some results:

image

**TODO:**
 - [x] tests
 - [x] docs
 - [x] example: https://colab.research.google.com/drive/12GdbLf4mB_jni1vNICDwXxNyigYKd1Y4?usp=sharing 
 - [ ] benchmarks
   - https://www.kaggle.com/nbuhagiar/carla-semantic-segmentation

## Before submitting
- [x] Was this discussed/approved via a Github issue? (no need for typos and docs improvements)
- [x] Did you read the [contributor guideline](https://github.com/PyTorchLightning/lightning-flash/tree/master/.github/CONTRIBUTING.md), Pull Request section?
- [x] Did you make sure your PR does only one thing, instead of bundling different changes together?
- [x] Did you make sure to update the documentation with your changes?
- [x] Did you write any new necessary tests? [not needed for typos/docs]
- [x] Did you verify new and existing tests pass locally with your changes?
- [ ] If you made a notable change (that affects users), did you update the [CHANGELOG](https://github.com/PyTorchLightning/lightning-flash/blob/master/CHANGELOG.md)?

## PR review
 - [x] Is this pull request ready for review? (if not, please submit in draft mode)

Anyone in the community is free to review the PR once the tests have passed.
If we didn't discuss your PR in Github issues there's a high chance it will not be merged.

## Did you have fun?
Make sure you had fun coding 🙃

PR #256, opened 2021-04-29
[1/N] Data Sources

ethanwharris:
## What does this PR do?

Adds data sources, docs and new tests to come in later PRs

## Before submitting

## PR review

## Did you have fun?


flash/core/model.py
```diff
@@ -110,7 +111,8 @@ def step(self, batch: Any, batch_idx: int) -> Any:
         """
         The training/validation/test step. Override for custom behavior.
         """
-        x, y = batch
+        x, y = batch['input'], batch['target']
```
edgarriba, line 114: enums here ? I see that you use it in some places

flash/data/data_source.py
```diff
@@ -0,0 +1,263 @@
...
+    @property
+    def preprocess(self) -> Optional[Preprocess]:
+        return self._preprocess
+
+    @abstractmethod
+    def load_data(
+        self,
+        data: Any,
+        dataset: Optional[Any] = None
+    ) -> Iterable[Mapping[str, Any]]:  # TODO: decide what type this should be
```
edgarriba, line 83: where does string suppose to point, specific files, images, ? It's a bit confusing what the `load_data` and `load_sample` methods are doing.

flash/data/data_source.py
```diff
@@ -0,0 +1,263 @@
...
+        if targets is None:
+            return [{'input': input} for input in inputs]
+        return [{'input': input, 'target': target} for input, target in zip(inputs, targets)]
+
+
+class FoldersDataSource(DataSource, ABC):
+
+    def __init__(
+        self,
+        train_folder: Optional[Union[str, pathlib.Path, list]] = None,
```
edgarriba, line 185: I vote to only support primitive types here `list` and `list(str)`

flash/data/data_source.py
```diff
@@ -0,0 +1,263 @@
...
+            train_data=train_folder,
+            val_data=val_folder,
+            test_data=test_folder,
+            predict_data=predict_folder,
+        )
+
+        self.extensions = extensions
+
+    @staticmethod
+    def find_classes(dir: str) -> Tuple[List[str], Dict[str, int]]:
```
edgarriba, line 201: shouldn't this be specific to classification tasks ?

flash/data/data_source.py
```diff
@@ -0,0 +1,263 @@
...
+        dataset.num_classes = len(classes)
+        data = make_dataset(data, class_to_idx, extensions=self.extensions)
+        return [{'input': input, 'target': target} for input, target in data]
+
+
+class FilesDataSource(DataSource, ABC):
+
+    def __init__(
+        self,
+        train_files: Optional[Union[Sequence[Union[str, pathlib.Path]], Iterable[Union[str, pathlib.Path]]]] = None,
```
edgarriba, line 237: same as before, I would simplify to primitive types the inputs

flash/data/data_source.py
```diff
@@ -0,0 +1,263 @@
...
+            train_data=(train_files, train_targets),
+            val_data=(val_files, val_targets),
+            test_data=(test_files, test_targets),
+            predict_data=(predict_files, None),
+        )
+
+        self.extensions = extensions
+
+    def load_data(self, data: Any, dataset: Optional[Any] = None) -> Iterable[Mapping[str, Any]]:
+        # TODO: Bring back the code to work out how many classes there are
```
edgarriba, line 256: `_load_data_from_tuple` , etc ?

flash/data/transforms.py
```diff
@@ -0,0 +1,28 @@
+from typing import Any, Mapping, Sequence, Union
+
+from torch import nn
+
+from flash.data.utils import convert_to_modules
+
+
+class ApplyToKeys(nn.Sequential):
```
edgarriba, line 8: this will work only for single inputs. E.g. in segmentation or tasks that require augmenting not only the inputs we still need something like this 
https://github.com/PyTorchLightning/lightning-flash/pull/239/files#diff-10dfb9297339fe1066151bc80b43feae338b80bfc185f47b951530e388b68719R38

flash/vision/data.py
```diff
@@ -0,0 +1,61 @@
...
+        super().__init__(
+            train_folder=train_folder,
+            val_folder=val_folder,
+            test_folder=test_folder,
+            predict_folder=predict_folder,
+            extensions=IMG_EXTENSIONS,
+        )
+
+    def load_sample(self, sample: Mapping[str, Any], dataset: Optional[Any] = None) -> Any:
+        result = {}  # TODO: this is required to avoid a memory leak, can we automate this?
```
edgarriba, line 28: also both `load_sample` are the same

flash/data/data_source.py
ethanwharris: Strings are the keys. So data can be any list of dicts like `{'input': ..., 'target': ...}`. So the `load_data` and `load_sample` are the same as before, just now have to be mappings rather than tuples, but I still need to add back the docstring :smiley:

flash/data/data_source.py
ethanwharris: Yeah, typing is wrong here as it can't be a list anyway, but we should maybe just do str

flash/data/data_source.py
ethanwharris: If it can't get the classes then it will just yield the files and no targets so would all still work fine. It gets a bit complicated if we factor this out into a `ClassificationFoldersDataSource` but we may need to, we'll know more once I add support for the other tasks.

flash/data/data_source.py
ethanwharris: Yeah, this should just be sequence of str

flash/data/data_source.py
ethanwharris: Not sure what you mean here, `data` should always be a tuple, just can sometimes be not a tuple if we are predicting.

flash/data/transforms.py
ethanwharris: It's tricky, this class allows multi-input by extracting mutliple keys to then be passed to a single aug. So if you had `my_augmentation(image=..., mask=...)` then this way will work. It's not quite right as it assumes tuple inputs (which won't work with albumenmtations) and we will need something different again for Kornia haha. So, overall, we need something smarter here, but I don't know what yet

flash/core/model.py
tchaton: Let's move this in the children classes.

```
def training_step
       self.step((batch['input'], batch['target']))
      ....
```

flash/vision/data.py
ethanwharris: Yeah, the issue is that each class needs to have different `load_data` methods but the same `load_sample`, I'm working on a way to clean this up haha

flash/data/transforms.py
edgarriba: agree, my only point was that to apply the same transformation to different inputs we might need to hold the generated random parameters.

flash/data/data_module.py
```diff
@@ -53,20 +56,35 @@ class DataModule(pl.LightningDataModule):
 
     def __init__(
         self,
-        train_dataset: Optional[Dataset] = None,
```
tchaton, line 56: It should still work if users provide their own dataset.

tchaton: Do we want to make this DataModule Flash specific ?

flash/data/data_pipeline.py
```diff
@@ -506,24 +506,24 @@ def _detach_postprocess_from_model(model: 'Task'):
...
-    def _generate_auto_dataset(
-        self,
-        data: Union[Iterable, Any],
-        running_stage: RunningStage = None,
-        use_iterable_auto_dataset: bool = False
-    ) -> Union[AutoDataset, IterableAutoDataset]:
-        if use_iterable_auto_dataset:
-            return IterableAutoDataset(data, data_pipeline=self, running_stage=running_stage)
-        return AutoDataset(data=data, data_pipeline=self, running_stage=running_stage)
+    # def _generate_callable_auto_dataset(
```
tchaton, line 509: Can be removed.

flash/data/data_source.py
```diff
@@ -0,0 +1,263 @@
...
+        test_dataset = self._generate_dataset_if_possible(RunningStage.TESTING, data_pipeline)
+        predict_dataset = self._generate_dataset_if_possible(RunningStage.PREDICTING, data_pipeline)
+        return train_dataset, val_dataset, test_dataset, predict_dataset
+
+    def _generate_dataset_if_possible(
+        self,
+        running_stage: RunningStage,
+        data_pipeline: DataPipeline,
+    ) -> Optional[Union[AutoDataset, IterableAutoDataset]]:
+        data = getattr(self, f"{_STAGES_PREFIX[running_stage]}_data", None)
```
tchaton, line 114: Let's support prefix there for `predict_load_data` there too.

flash/data/data_source.py
```diff
@@ -0,0 +1,263 @@
...
+        if data is not None:
+            return self.generate_dataset(data, running_stage, data_pipeline)
+
+    def generate_dataset(
+        self,
+        data,
+        running_stage: RunningStage,
+        data_pipeline: DataPipeline,
+    ) -> Optional[Union[AutoDataset, IterableAutoDataset]]:
+        mock_dataset = MockDataset()
```
tchaton, line 124: Nice hack :)

flash/data/data_source.py
```diff
@@ -0,0 +1,263 @@
...
+        super().__init__(
+            train_data=(train_files, train_targets),
+            val_data=(val_files, val_targets),
+            test_data=(test_files, test_targets),
+            predict_data=(predict_files, None),
+        )
+
+        self.extensions = extensions
+
+    def load_data(self, data: Any, dataset: Optional[Any] = None) -> Iterable[Mapping[str, Any]]:
```
tchaton, line 255: I would differentiate predict_load_data to make it slightly cleaner.

flash/data/data_source.py
tchaton: what is the difference with `SequenceDataSource` class ?

flash/vision/classification/data.py
```diff
@@ -178,111 +164,18 @@ def _resolve_transforms(
...
+        return default_collate(samples)
 
     def common_step(self, sample: Any) -> Any:
-        if isinstance(sample, (list, tuple)):
-            source, target = sample
-            return self.current_transform(source), target
+        # if isinstance(sample, (list, tuple)):
+        #     source, target = sample
+        #     return self.current_transform(source), target
         return self.current_transform(sample)
```
tchaton, line 179: Let's move this to the base class in the next PR :)

flash/vision/classification/data.py
```diff
@@ -178,111 +164,18 @@ def _resolve_transforms(
...
+            for key in sample.keys():
+                if torch.is_tensor(sample[key]):
+                    sample[key] = sample[key].squeeze(0)
+        return default_collate(samples)
 
     def common_step(self, sample: Any) -> Any:
-        if isinstance(sample, (list, tuple)):
-            source, target = sample
-            return self.current_transform(source), target
+        # if isinstance(sample, (list, tuple)):
```
tchaton, line 176: Extra comments.

flash/vision/classification/data.py
```diff
@@ -178,111 +164,18 @@ def _resolve_transforms(
...
+        #     source, target = sample
+        #     return self.current_transform(source), target
```
tchaton, line 178: Can we clean to_tensor_transform too ?

flash/data/data_module.py
```diff
@@ -53,20 +56,35 @@ class DataModule(pl.LightningDataModule):
...
+
+        self._preprocess: Optional[Preprocess] = preprocess
+        self._postprocess: Optional[Postprocess] = postprocess
+        self._viz: Optional[BaseVisualization] = None
+        self._data_fetcher: Optional[BaseDataFetcher] = data_fetcher or self.configure_data_fetcher()
+
+        # TODO: Preprocess can change
+        self.data_fetcher.attach_to_preprocess(self.preprocess)
+
+        train_dataset, val_dataset, test_dataset, predict_dataset = data_source.to_datasets(self.data_pipeline)
```
tchaton, line 78: This line is confusing me. Why do we provide the data_pipeline ?

tchaton (commented):
Great PR. Missing tests :)

flash/data/data_module.py
ethanwharris: Done :)

flash/data/data_source.py
ethanwharris: Done

flash/data/data_module.py
edgarriba: I vote yes, otherwise can be confusing between lightning/flash components and their namings

flash/data/data_module.py
```diff
@@ -367,113 +332,115 @@ def _split_train_val(
...
         data_fetcher: BaseDataFetcher = None,
         preprocess: Optional[Preprocess] = None,
-        postprocess: Optional[Postprocess] = None,
-        use_iterable_auto_dataset: bool = False,
-        seed: int = 42,
         val_split: Optional[float] = None,
-        **kwargs,
+        batch_size: int = 4,
+        num_workers: Optional[int] = None,
+        **kwargs: Any,
```
edgarriba, line 382: what would this `kwargs` be ?

flash/data/data_module.py
```diff
@@ -367,113 +332,115 @@ def _split_train_val(
...
+            batch_size=batch_size,
+            num_workers=num_workers,
         )
 
-        if train_dataset is not None and (val_split is not None and val_dataset is None):
-            train_dataset, val_dataset = cls._split_train_val(train_dataset, val_split)
+    @classmethod
+    def from_files(
+        cls,
+        train_files: Optional[Sequence[Union[str, pathlib.Path]]] = None,
```
edgarriba, line 407: I would standardize just to accept `str`

flash/data/data_module.py
ethanwharris: Have semi-reverted this so datamodule can now be made from datasets again

tests/data/test_auto_dataset.py
```diff
@@ -137,13 +217,15 @@ def test_autodataset_with_functions(
         assert functions.load_sample_count == len(dset)
 
 
+# TODO: do we remove ?
```
edgarriba, line 220: @ethanwharris check this tests is make sense to keep it

tests/data/test_auto_dataset.py
```diff
@@ -137,13 +217,15 @@ def test_autodataset_with_functions(
...
 def test_autodataset_warning():
     with pytest.warns(
         UserWarning, match="``datapipeline`` is specified but load_sample and/or load_data are also specified"
     ):
         AutoDataset(range(10), load_data=lambda x: x, load_sample=lambda x: x, data_pipeline=DataPipeline())
 
 
+# TODO: do we remove ?
```
edgarriba, line 228: @ethanwharris check this tests is make sense to keep it

tests/data/test_auto_dataset.py
```diff
@@ -173,6 +255,7 @@ def test_preprocessing_data_pipeline_with_running_stage(with_dataset):
         assert pipe._preprocess_pipeline.train_load_data_count == 1
 
 
+# TODO: do we remove ? we are testing DataPipeline here.
```
edgarriba, line 258: @ethanwharris check this tests is make sense to keep it

flash/data/data_source.py
edgarriba: add a todo here to move this later

flash/data/data_source.py
```diff
@@ -0,0 +1,264 @@
...
+        classes = [d.name for d in os.scandir(dir) if d.is_dir()]
+        classes.sort()
+        class_to_idx = {cls_name: i for i, cls_name in enumerate(classes)}
+        return classes, class_to_idx
+
+    @staticmethod
+    def isdir(data: Union[str, Tuple[List[str], List[Any]]]) -> bool:
+        try:
+            return os.path.isdir(data)
+        except TypeError:
```
edgarriba, line 217: in which case this will fail ?

flash/data/utils.py
```diff
@@ -32,9 +32,13 @@
 }
 _STAGES_PREFIX_VALUES = {"train", "test", "val", "predict"}
 
+_DATASOURCE_FUNCS: Set[str] = {
+    'load_data',
```
edgarriba, line 36:
```suggestion
    "load_data",
```

flash/data/utils.py
```diff
@@ -32,9 +32,13 @@
...
+    'load_sample',
```
edgarriba, line 37:
```suggestion
    "load_sample",
```

flash/tabular/classification/data/data.py
```diff
@@ -58,210 +54,142 @@ def __init__(
...
+    ):
         # impute_data
         # compute train dataset stats
         dfs = _pre_transform([df], self.num_cols, self.cat_cols, self.codes, self.mean, self.std, self.target_col,
                              self.target_codes)
 
         df = dfs[0]
 
-        dataset.num_samples = len(df)
+        if dataset is not None:
```
edgarriba, line 74:
```suggestion
        if dataset is not None and hasattr(dataset, 'num_samples'):
```

flash/tabular/classification/data/data.py
```diff
@@ -58,210 +54,142 @@ def __init__(
...
-        cat_cols: List[str],
-        is_regression: bool,
-    ) -> 'TabularPreprocess':
 
-        if train_df is None:
-            raise MisconfigurationException("train_df is required to instantiate the TabularPreprocess")
+class TabularCSVDataSource(TabularDataFrameDataSource):
 
-        dfs = [train_df]
+    def load_data(self, data: str, dataset: Optional[Any] = None):
```
edgarriba, line 99: not sure if we should check somehow if the file has a valid extension

flash/tabular/classification/data/data.py
```diff
@@ -342,25 +287,58 @@ def from_df(
...
+            train_df, val_df, test_df, predict_df, target_col, numerical_cols, categorical_cols
         )
 
-        return cls.from_load_data_inputs(
-            train_load_data_input=train_df,
-            val_load_data_input=val_df,
-            test_load_data_input=test_df,
-            predict_load_data_input=predict_df,
+        return cls.from_data_source(
+            data_source="df",
```
edgarriba, line 295: I would for a more explicit naming "data_frame"

flash_examples/custom_task.py
```diff
@@ -77,12 +84,13 @@ def from_dataset(cls, x: ND, y: ND, preprocess: Preprocess, batch_size: int = 64
 
         x_train, x_test, y_train, y_test = train_test_split(x, y, test_size=.20, random_state=0)
 
-        dm = cls.from_load_data_inputs(
-            train_load_data_input=(x_train, y_train),
-            test_load_data_input=(x_test, y_test),
+        dm = cls.from_data_source(
+            "numpy",
```
edgarriba, line 88: @ethanwharris now seeing this the example - I see a bit redundant to specify in the constructor the data source type and again here

flash/tabular/classification/model.py
```diff
@@ -80,9 +81,22 @@ def __init__(
     def forward(self, x_in) -> torch.Tensor:
         # TabNet takes single input, x_in is composed of (categorical, numerical)
         x = torch.cat([x for x in x_in if x.numel()], dim=1)
-        return F.softmax(self.model(x)[0], -1)
+        return self.model(x)[0]
+
+    def training_step(self, batch: Any, batch_idx: int) -> Any:
+        batch = (batch[DefaultDataKeys.INPUT], batch[DefaultDataKeys.TARGET])
```
tchaton, line 87: Neat !

flash/text/classification/data.py
```diff
@@ -13,138 +13,63 @@
...
-            else:
-                raise MisconfigurationException(
-                    "Either ``label_to_class_mapping`` or ``train_file`` needs to be provided"
-                )
-
-        self.backbone = backbone
-        self.tokenizer = AutoTokenizer.from_pretrained(backbone, use_fast=True)
-        self.input = input
-        self.filetype = filetype
+        self.tokenizer = tokenizer
```
tchaton, line 35: Better to create the tokenizer in the DataSource. It is simpler to pickle.

flash/text/classification/data.py
```diff
@@ -155,37 +80,113 @@ def load_data(
...
+    def per_batch_transform(self, batch: Any) -> Any:
+        if "labels" not in batch:
+            # todo: understand why an extra dimension has been added.
+            if batch["input_ids"].dim() == 3:
+                batch["input_ids"] = batch["input_ids"].squeeze(0)
+        return batch
+
+    def collate(self, samples: Any) -> Tensor:
+        """Override to convert a set of samples to a batch"""
+        if isinstance(samples, dict):
```
tchaton, line 187: same here ?

flash/video/classification/data.py
```diff
@@ -265,19 +140,68 @@ def from_paths(
...
+            'clip_sampler': self.clip_sampler,
+            'clip_duration': self.clip_duration,
+            'clip_sampler_kwargs': self.clip_sampler_kwargs,
+            'video_sampler': self.video_sampler,
+            'decode_audio': self.decode_audio,
+            'decoder': self.decoder,
+        }
+
+    @classmethod
+    def load_state_dict(cls, state_dict: Dict[str, Any], strict: bool) -> 'VideoClassificationPreprocess':
```
tchaton, line 174: Can we move this to the base class ?

flash/video/classification/data.py
```diff
@@ -265,19 +140,68 @@ def from_paths(
...
-            predict_load_data_input=predict_data_path,
-            batch_size=batch_size,
-            num_workers=num_workers,
-            preprocess=preprocess,
-            use_iterable_auto_dataset=True,
-            **kwargs,
-        )
+    def get_state_dict(self) -> Dict[str, Any]:
+        return {
+            'train_transform': self._train_transform,
```
tchaton, line 161: I think we should swap between `_train_transform` and `train_transform` in the preprocess internals.
The public should be the provided by the user.

edgarriba (approved):
LGTM. Very few comments and questions

flash/vision/data.py
```diff
@@ -0,0 +1,44 @@
...
+
+    def load_sample(self, sample: Dict[str, Any], dataset: Optional[Any] = None) -> Dict[str, Any]:
+        sample[DefaultDataKeys.INPUT] = to_pil_image(sample[DefaultDataKeys.INPUT])
+        return sample
+
+
+class ImageNumpyDataSource(NumpyDataSource):
+
+    def load_sample(self, sample: Dict[str, Any], dataset: Optional[Any] = None) -> Dict[str, Any]:
+        sample[DefaultDataKeys.INPUT] = to_pil_image(torch.from_numpy(sample[DefaultDataKeys.INPUT]))
```
tchaton, line 43: Does this make sense ? We are still trying to get something optimised. Are we converting from NumPy to torch to pil to torch ?

tchaton: Let's adapt to_tensor_tensor to work on NumPy array directly.

ethanwharris: My thinking here was that the augmentations (which expect PIL images) should still work from numpy and from tensor. This way the behaviour is identical whether you use from_numpy or from_folders. If we change it then the defaults will give an error with numpy

flash/tabular/classification/data/data.py
ethanwharris: We are setting the attribute here, so it doesn't need to have it beforehand

flash/tabular/classification/data/data.py
ethanwharris: I guess that pandas will do this for us and give a reasonable error if it's not a csv

flash/core/model.py
```diff
@@ -154,6 +158,7 @@ def test_step(self, batch: Any, batch_idx: int) -> None:
     def predict(
         self,
         x: Any,
+        data_source: Union[str] = "default",
```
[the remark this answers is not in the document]
ethanwharris: Fixed :)

flash/data/data_module.py
```diff
@@ -367,113 +334,263 @@ def _split_train_val(
...
+            num_workers=num_workers,
+            **preprocess_kwargs,
         )
 
-        if train_dataset is not None and (val_split is not None and val_dataset is None):
-            train_dataset, val_dataset = cls._split_train_val(train_dataset, val_split)
+    @classmethod
+    def from_json(
+        cls,
+        input_fields: Union[str, List[str]],
```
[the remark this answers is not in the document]
ethanwharris: Have gone for Sequence

flash/data/data_source.py
```diff
@@ -0,0 +1,264 @@
...
+
+    def generate_dataset(
+        self,
+        data: Optional[DATA_TYPE],
+        running_stage: RunningStage,
+    ) -> Optional[Union[AutoDataset, IterableAutoDataset]]:
+        is_none = data is None
+        # TODO: we should parse better the possible data types here.
+        # Are `pata_paths` considered as Sequence ? for now it pass
+        # the statement found in below.
```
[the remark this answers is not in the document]
ethanwharris: Done

flash/data/data_source.py
```diff
@@ -0,0 +1,264 @@
...
+        if isinstance(data, Sequence):
+            is_none = data[0] is None
+
+        if not is_none:
+            from flash.data.data_pipeline import DataPipeline
+
+            mock_dataset = MockDataset()
+            with CurrentRunningStageFuncContext(running_stage, "load_data", self):
+                load_data = getattr(
```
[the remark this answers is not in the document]
ethanwharris: Done

flash/data/data_source.py
ethanwharris: I've added a comment explaining, but is just duck typing in case it is a list of paths

flash/tabular/classification/data/data.py
ethanwharris: Yeah, `data_frame` is better, cahnged it everywhere

flash/core/model.py
```diff
@@ -169,9 +174,10 @@ def predict(
         """
         running_stage = RunningStage.PREDICTING
 
-        data_pipeline = self.build_data_pipeline(data_pipeline)
+        data_pipeline = self.build_data_pipeline(data_source, data_pipeline)
 
-        x = [x for x in data_pipeline._generate_auto_dataset(x, running_stage)]
+        x = [x for x in data_pipeline._data_source.generate_dataset(x, running_stage)]
```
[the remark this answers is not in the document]
ethanwharris: The `data_source` passed to `predict` is just a string, only becomes a usable object once it is resolved and given to the data pipeline. We could make it a public attribute of data_pipeline?

flash/data/auto_dataset.py
```diff
@@ -38,141 +37,68 @@ class BaseAutoDataset:
 
     def __init__(
         self,
-        data: Any,
-        load_data: Optional[Callable] = None,
-        load_sample: Optional[Callable] = None,
-        data_pipeline: Optional['DataPipeline'] = None,
-        running_stage: Optional[RunningStage] = None
+        data: DATA_TYPE,
```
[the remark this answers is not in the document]
ethanwharris: Done :)

flash/data/process.py
```diff
@@ -338,6 +253,33 @@ def _save_to_state_dict(self, destination, prefix, keep_vars):
...
+
+    def _resolve_transforms(
+        self,
+        train_transform: Optional[Union[str, Dict]] = 'default',
+        val_transform: Optional[Union[str, Dict]] = 'default',
+        test_transform: Optional[Union[str, Dict]] = 'default',
+        predict_transform: Optional[Union[str, Dict]] = 'default',
+    ):
+        if not train_transform or train_transform == 'default':
+            train_transform = self.default_train_transforms()
```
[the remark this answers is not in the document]
ethanwharris: Updated :)

justusschock (commented):
I like it. So far I just don't get the following: If I want to implement a new task with a completely new datatype: What and how do I have to implement it?

flash/text/classification/data.py
ethanwharris: Done :smiley:

flash/core/model.py
```diff
@@ -305,8 +320,16 @@ def build_data_pipeline(self, data_pipeline: Optional[DataPipeline] = None) -> O
                 getattr(data_pipeline, '_serializer', None),
             )
 
-        data_pipeline = DataPipeline(preprocess, postprocess, serializer)
-        data_pipeline.initialize()
+        data_source = data_source or old_data_source
+
+        if isinstance(data_source, str):
+            if preprocess is None:
+                data_source = DataSource()  # TODO: warn the user that we are not using the specified data source
```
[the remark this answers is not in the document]
ethanwharris: I have some ideas for a data sources registry. Not sure it makes sense to just map strings to data sources as most data sources only work with particular preprocesses.

flash/data/auto_dataset.py
```diff
@@ -12,22 +12,21 @@
...
 from torch.utils.data import Dataset, IterableDataset
 
-from flash.data.callback import ControlFlow
-from flash.data.process import Preprocess
-from flash.data.utils import _STAGES_PREFIX, _STAGES_PREFIX_VALUES, CurrentRunningStageFuncContext
+from flash.data.utils import CurrentRunningStageFuncContext
 
 if TYPE_CHECKING:
     from flash.data.data_pipeline import DataPipeline
+    from flash.data.data_source import DataSource
```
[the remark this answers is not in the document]
ethanwharris: Docs build is working for now, I'm not sure we can avoid a circular import here but could maybe just import the module and type as `data_source.DataSource`.

flash/data/data_module.py
```diff
@@ -367,113 +333,263 @@ def _split_train_val(
...
+    @classmethod
+    def from_json(
+        cls,
+        input_fields: Union[str, Sequence[str]],
+        target_fields: Optional[Union[str, Sequence[str]]] = None,
+        train_file: Optional[str] = None,
+        val_file: Optional[str] = None,
+        test_file: Optional[str] = None,
+        predict_file: Optional[str] = None,
+        data_fetcher: BaseDataFetcher = None,
```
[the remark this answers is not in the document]
ethanwharris: Fixed :)

flash/data/data_module.py
```diff
@@ -367,113 +333,263 @@ def _split_train_val(
...
-            **kwargs
+    @classmethod
+    def from_csv(
+        cls,
+        input_fields: Union[str, Sequence[str]],
+        target_fields: Optional[Union[str, Sequence[str]]] = None,
+        train_file: Optional[str] = None,
+        val_file: Optional[str] = None,
+        test_file: Optional[str] = None,
+        predict_file: Optional[str] = None,
```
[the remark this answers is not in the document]
ethanwharris: Fixed :)

flash/data/data_source.py
```diff
@@ -0,0 +1,263 @@
...
+    ) -> Optional[Union[AutoDataset, IterableAutoDataset]]:
+        is_none = data is None
+
+        if isinstance(data, Sequence):
+            is_none = data[0] is None
+
+        if not is_none:
+            from flash.data.data_pipeline import DataPipeline
+
+            mock_dataset = MockDataset()
```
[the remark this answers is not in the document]
ethanwharris: Need a mock dataset because the actual dataset object doesn't exist yet. Later on this can probably be refactored with some thought to just use the state system or something like that, but fo now I have used typing.cast :smiley:

flash/data/data_source.py
```diff
@@ -0,0 +1,263 @@
...
+
+    PATHS = "paths"
+    NUMPY = "numpy"
+    TENSOR = "tensor"
+    CSV = "csv"
+    JSON = "json"
+
+    # TODO: Create a FlashEnum class???
+    def __hash__(self) -> int:
+        return hash(self.value)
```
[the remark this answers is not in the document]
ethanwharris: Not sure, the hash in lightning enum uses the name for some reason. If that is not deliberate then we can upstream this to lightning, otherwise we could make a flashenum class that does what we need.

flash/data/callback.py
```diff
@@ -156,7 +156,8 @@ def __init__(self, enabled: bool = False):
     def _store(self, data: Any, fn_name: str, running_stage: RunningStage) -> None:
         if self.enabled:
             store = self.batches[_STAGES_PREFIX[running_stage]]
-            store.setdefault(fn_name, [])
+            if fn_name not in store:
+                store[fn_name] = []
```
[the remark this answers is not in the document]
ethanwharris: Whoops, I changed that when debugging a weird issue that turned out to be something else, reverted :)

flash/core/model.py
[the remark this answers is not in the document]
ethanwharris: Done :smiley:

flash/video/classification/data.py
ethanwharris: Done :smiley:

flash/video/classification/data.py
ethanwharris: If we move to the base class then the user would no longer be required to implement `get_state_dict` and `load_state_dict` themselves. I played with a few options, best I could come up with was to add a `transforms` property so that it's just one line.

ethanwharris:
@edgarriba @tchaton @justusschock Thanks for all the feedback, I've now finished responding to comments, let me know if you have anything further.

@justusschock Regarding adding new tasks with custom data, the user would just implement their own `DataSource` and set this up in their preprocess. They can then either add a custom `from_*` method to their `DataModule` or just use the `from_data_source` method. An example of this is the "coco" data source in `/vision/detection/data`

tchaton (approved):
LGTM !

flash/core/model.py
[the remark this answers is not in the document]
justusschock: Yes, but the registry would most likely be task specific as well (similar to backbones).

flash/data/data_source.py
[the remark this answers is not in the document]
justusschock: @tchaton thoughts?

PR #262, opened 2021-05-05
[WIP] add style transfer task with pystiche

pmeier:
## What does this PR do?

Add a style transfer task using [`pystiche`](https://pystiche.org) as backend.

Note: Change codeblock to test-code when 0.7.2 is out.  

## Before submitting
- [x] Was this discussed/approved via a Github issue? (no need for typos and docs improvements) pystiche/pystiche#484

## PR review

## Did you have fun?


flash/vision/style_transfer/data.py
```diff
@@ -0,0 +1 @@
+# TODO
```
pmeier, line 1: I need an image dataset preferably unsupervised here. The original model was trained on the COCO 2014 train split.

flash/vision/style_transfer/model.py
```diff
@@ -0,0 +1,90 @@
...
+        super().__init__()
+        self.conv1 = Conv(channels, channels, kernel_size=3)
+        self.conv2 = Conv(channels, channels, kernel_size=3, activation=False)
+
+    def forward(self, input):
+        output = self.conv2(self.conv1(input))
+        return output + input
+
+
+class Transformer(nn.Module):
```
pmeier, line 70: This is the model that should be trained. It has no canonical name. This is compatible with the [PyTorch example](https://github.com/pytorch/examples/tree/master/fast_neural_style) which in turn follows [the initial paper](https://link.springer.com/content/pdf/10.1007%2F978-3-319-46475-6_43.pdf).

flash_examples/predict/style_transfer.py
```diff
@@ -0,0 +1,47 @@
...
+    multi_layer_encoder,
+    style_layers,
+    lambda encoder, layer_weight: GramOperator(encoder, score_weight=layer_weight),
+    layer_weights="sum",
+    score_weight=style_weight,
+)
+
+# TODO: this needs to be moved to the device to be trained on
+# TODO: we need to register a style image here
+perceptual_loss = loss.PerceptualLoss(content_loss, style_loss)
```
pmeier, line 42: I'm not sure if this should go here or not. The perceptual loss together with the model are the core parts. So if the user wants to change things it is most likely one of the two.

flash_examples/predict/style_transfer.py
```diff
@@ -0,0 +1,47 @@
...
+content_loss = ops.FeatureReconstructionOperator(
+    content_encoder, score_weight=content_weight
+)
+
+
+class GramOperator(ops.GramOperator):
+    def enc_to_repr(self, enc: torch.Tensor) -> torch.Tensor:
+        repr = super().enc_to_repr(enc)
+        num_channels = repr.size()[1]
+        return repr / num_channels
```
pmeier, line 27: This change is only done to be compatible with the PyTorch example / paper. They both implement a double normalization by the number of channels which is most likely a bug. We could remove it here, but we would need to tune the hyperparameters then.

pmeier (commented):
This is just the very preliminary state. I've hit several roadblocks that need to be resolved:

- Models for Neural Style Transfer are trained in an unsupervised manner. Thus, I simply need a dataset that can supply images without labels / annotations.
- Following from the point above, there is no train / val / test split. The process is done after the training. If there is something like a validation / test it is performed manually by trying a few examples. There is no objective way to put a number on the quality of the stylization.
- The models used as transformer are not named. I've seen that the model is usually loaded by their name, which is thus not possible. We could fall back to a author / year combination of the paper the architecture was published. 

I'll fix the linting errors and update the documentation, tests, and the changelog when the main part is resolved.

flash/vision/__init__.py
```diff
@@ -2,3 +2,4 @@
 from flash.vision.classification import ImageClassificationData, ImageClassificationPreprocess, ImageClassifier
 from flash.vision.detection import ObjectDetectionData, ObjectDetector
 from flash.vision.embedding import ImageEmbedder
+from .style_transfer import *
```
tchaton, line 5: use direct import everywhere :)

flash/vision/style_transfer/model.py
```diff
@@ -0,0 +1,90 @@
+from torch import nn
+from torch.nn.functional import interpolate
+
+__all__ = ["Transformer"]
+
+
+class Interpolate(nn.Module):
```
tchaton, line 7: Could this model be imported directly from `pystiche`. The goal of Flash is to rely on other libraries to enable tasks, not implement nn layers :)

flash_examples/predict/style_transfer.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+style_layers = ("relu1_2", "relu2_2", "relu3_3", "relu4_3")
+style_weight = 1e10
+style_loss = ops.MultiLayerEncodingOperator(
```
tchaton, line 32: Everything should be included in a Flash Task.

tchaton:
Hey @pmeier,

Awesome you started.

```
class StyleTransfer(Task):

    models: FlashRegistry = STYLE_TRANSFER_MODELS

    def __init__(
        self,

        content_image: Union[Image.PIL, str, np.ndarray],
        style_loss: Optional[Callable] = None,
        content_loss: Optional[Callable] = None,
        perceptual_loss: Optional[Callable] = None,

        model: Union[str, Tuple[nn.Module, int]] = "transformer",
        model_kwargs: Optional[Dict] = None,
        optimizer: Union[Type[torch.optim.Optimizer], torch.optim.Optimizer] = torch.optim.Adam,
        optimizer_kwargs: Optional[Dict[str, Any]] = None,
        scheduler: Optional[Union[Type[_LRScheduler], str, _LRScheduler]] = None,
        scheduler_kwargs: Optional[Dict[str, Any]] = None,
        metrics: Union[torchmetrics.Metric, Mapping, Sequence, None] = None,
        learning_rate: float = 1e-3,
        serializer: Optional[Union[Serializer, Mapping[str, Serializer]]] = None,
    ):

        if perceptual_loss is None:
            content_loss = content_loss or self.default_content_loss()
            style_loss = style_loss or self.default_style_transfer()
            perceptual_loss = loss.PerceptualLoss(content_loss, style_loss)

        if content_image is not None:
            perceptual_loss.set_content_image(content_image)

        self.perceptual_loss = perceptual_loss

        self.save_hyperparameters()

        if isinstance(model, tuple):
            model = model
        else:
            model = self.models.get(model)(pretrained=pretrained, **model_kwargs)


        super().__init__(
            model=model,
            loss_fn=perceptual_loss,
            optimizer=optimizer,
            optimizer_kwargs=optimizer_kwargs,
            scheduler=scheduler,
            scheduler_kwargs=scheduler_kwargs,
            metrics=metrics,
            learning_rate=learning_rate,
            serializer=serializer,
        )


    def default_content_loss(self):
        multi_layer_encoder = enc.vgg16_multi_layer_encoder()
        content_layer = "relu2_2"
        content_encoder = multi_layer_encoder.extract_encoder(content_layer)
        content_weight = 1e5
        return = ops.FeatureReconstructionOperator(
            content_encoder, score_weight=content_weight
        )


    def default_style_transfer(self):
        class GramOperator(ops.GramOperator):
            def enc_to_repr(self, enc: torch.Tensor) -> torch.Tensor:
                repr = super().enc_to_repr(enc)
                num_channels = repr.size()[1]
                return repr / num_channels


        style_layers = ("relu1_2", "relu2_2", "relu3_3", "relu4_3")
        style_weight = 1e10
        return ops.MultiLayerEncodingOperator(
            multi_layer_encoder,
            style_layers,
            lambda encoder, layer_weight: GramOperator(encoder, score_weight=layer_weight),
            layer_weights="sum",
            score_weight=style_weight,
        )

    def forward(self, x):
        # not sure about this part
        self.model(x)
        return self.perceptual_loss(x)



# in finetuning.

content_image = ...
dm = StyleDataModule.from_folder(...)
model = StyleTransfer(content_image=content_image)
trainer = Trainer(...)
trainer.fit(model, dm)
```

flash/vision/style_transfer/data.py
edgarriba: @pmeier consider to look at https://github.com/PyTorchLightning/lightning-flash/pull/256 since the api for loading data will change quite a bit

flash/vision/style_transfer/model.py
edgarriba: as @tchaton mentioned, this model should be something that should be imported from `pystiche` and integrated it to the template task example that he provided

flash/vision/style_transfer/model.py
edgarriba: or use the `Resize`/`Rescale` module with those options that you implemented in `kornia` :)

flash_examples/predict/style_transfer.py
edgarriba: from flash perspective I believe this is something that should happen internally in your library. /cc @tchaton

flash/vision/style_transfer/model.py
pmeier: I'll leave it here for now. I need to push a release that includes this in `pystiche`. We shouldn't merge this PR until this is done.

flash/vision/style_transfer/model.py
```diff
@@ -0,0 +1,185 @@
...
+        if content_loss is None:
+            content_loss = self.default_content_loss(multi_layer_encoder)
+
+        if style_loss is None:
+            style_loss = self.default_style_loss(multi_layer_encoder)
+
+        self.perceptual_loss = loss.PerceptualLoss(content_loss, style_loss)
+        self.perceptual_loss.set_style_image(style_image)
+
+        self.save_hyperparameters()
```
pmeier, line 132: Couldn't find documentation for this. Is this needed? If yes, what does it do?

flash/vision/style_transfer/model.py
```diff
@@ -0,0 +1,185 @@
...
+
+    def forward(self, input: torch.Tensor) -> torch.Tensor:
+        return self.decoder(self.encoder(input))
+
+
+class StyleTransfer(Task):
+
+    def __init__(
+        self,
+        style_image: torch.Tensor,
```
pmeier, line 107: `pystiche` has built-in support for reading images from path. If you want support for `PIL.Image.Image` and `np.ndarray` I need some functionality to transform them into a tensor. `torchvision` has [`transforms.ToTensor`](https://pytorch.org/vision/stable/transforms.html#torchvision.transforms.ToTensor) for that.

flash/vision/style_transfer/model.py
```diff
@@ -0,0 +1,185 @@
...
+        model: Optional[nn.Module] = None,
+        multi_layer_encoder: Optional = None,
+        content_loss: Optional[Union[ops.ComparisonOperator, ops.OperatorContainer]] = None,
+        style_loss: Optional[Union[ops.ComparisonOperator, ops.OperatorContainer]] = None,
+        optimizer: Union[Type[torch.optim.Optimizer], torch.optim.Optimizer] = torch.optim.Adam,
+        optimizer_kwargs: Optional[Dict[str, Any]] = None,
+        scheduler: Optional[Union[Type[_LRScheduler], str, _LRScheduler]] = None,
+        scheduler_kwargs: Optional[Dict[str, Any]] = None,
+        metrics: Union[torchmetrics.Metric, Mapping, Sequence, None] = None,
```
pmeier, line 116: I see no point in accepting this, since there are no metrics for style transfer.

flash/vision/style_transfer/model.py
```diff
@@ -0,0 +1,185 @@
...
+        learning_rate: float = 1e-3,
+        serializer: Optional[Union[Serializer, Mapping[str, Serializer]]] = None,
```
pmeier, line 118: I didn't find any documentation on this. At least it is not explained in the docstring of `flash.Task`. I'm guessing this can stay, but IDK.

flash_examples/predict/style_transfer.py
```diff
@@ -0,0 +1,23 @@
...
+
+if _PYSTICHE_AVAILABLE:
+    import pystiche.demo
+else:
+    print("Please, run `pip install pystiche`")
+    sys.exit(0)
+
+download_data("http://images.cocodataset.org/zips/train2014.zip", "data")
+
+data_module = ImageUnsupervisedData.from_folder("data")
```
pmeier, line 16: I need something like a `ImageUnsupervised` class. If that is not possible, we could also use any other data loading class that returns all images. In that case, I need to know how to discard the class labels or annotations.

flash_examples/predict/style_transfer.py
```diff
@@ -0,0 +1,23 @@
...
+
+style_image = pystiche.demo.images()["paint"].read(size=256, edge="long")
+
+model = StyleTransfer(style_image)
+
+trainer = flash.Trainer(max_epochs=2)
```
pmeier, line 22: I also need `batch_size=4` but IDK where to put it. It doesn't seem to belong to the trainer.

pmeier (commented):
Apart from my comments / questions below, I'm wondering whether this example should be in `predict` or `finetune`.

flash/vision/style_transfer/model.py
edgarriba: ok - then please mark it as Draft

flash/vision/style_transfer/model.py
edgarriba: @tchaton @ethanwharris  is this somethings that we could automate inside the main `Task` class ?

flash/vision/style_transfer/model.py
edgarriba: if you already support torchvision internally - you could somehow expose or implement that internally

flash/vision/style_transfer/model.py
edgarriba: I would keep it for consistency with others task and maybe raise an error with a meaningful message. We don't know in future if anyone will come with goof metric for this task

flash_examples/predict/style_transfer.py
edgarriba: Flash Trainer sub-classes PL-Trainer which has `batch_size` as a member variable

flash/vision/style_transfer/model.py
edgarriba: Besides, what do you look at when you train style transfer - just the loss ?

flash_examples/predict/style_transfer.py
edgarriba: what are usually the inputs/labels for style transfer - pair of images ? We are about to introduce a similar data loading api for semantic segmentation that might be extend to other image generation tasks. /cc @ethanwharris 

@pmeier https://github.com/PyTorchLightning/lightning-flash/pull/239/files#diff-fff239f44797f0092593a14bf1af9a4b97d1959cba4a6a759eadbf126a697fc5R41

flash/vision/style_transfer/model.py
edgarriba: this is a relatively new flash component that helps to retrieve and serialize the expected results for a task. E.g. apply correct mapping to classification tasks. My guess is that in general image generation this just needs to return/save the final image. /cc @ethanwharris

flash_examples/predict/style_transfer.py
pmeier: > what are usually the inputs/labels for style transfer - pair of images ?

Style transfer is unsupervised, so there are not pairs. I just need the images.

flash_examples/predict/style_transfer.py
pmeier: There is no mention of a `batch_size` in the [docs](https://pytorch-lightning.readthedocs.io/en/stable/api/pytorch_lightning.trainer.trainer.html#trainer).

ethanwharris: `batch_size` is passed to the `DataModule` in the `from_*` method, not the trainer. So you should be able to do: `from_folders(..., batch_size=4)` :smiley:

flash/vision/style_transfer/model.py
pmeier: Is it possible to review draft PRs?

flash/vision/style_transfer/model.py
ethanwharris: Yeah, you may want a serializer that converts the style transferred tensor back into a PIL image to then be viewed.

flash/vision/style_transfer/model.py
ethanwharris: Yeah, I would remove it then just add it back if one day there is a metric

flash/vision/style_transfer/model.py
ethanwharris: You should be able to configure the to tensor transform in the default_transforms in the preprocess

flash/vision/style_transfer/model.py
```diff
@@ -0,0 +1,210 @@
...
+            lambda encoder, layer_weight: GramOperator(encoder, score_weight=layer_weight),
+            layer_weights="sum",
+            score_weight=style_weight,
+        )
+
+    def forward(self, content_image: torch.Tensor) -> torch.Tensor:
+        self.perceptual_loss.set_content_image(content_image)
+        return self.model(content_image)
+
+    def validation_step(self, batch: Any, batch_idx: int) -> NoReturn:
```
ethanwharris, line 206: Should also override `train_step` and `predict_step` here to get the `DefaultDataKeys.INPUT` from the batch. Example here: https://github.com/PyTorchLightning/lightning-flash/blob/d5ca5e526ff1cb3bad23a48398c4634b7fa37724/flash/vision/classification/model.py#L110

ethanwharris (commented):
It's getting there, just a couple of comments :smiley:

pmeier:
Training is running locally at the moment. Afterwards I'll update the documentation and the changelog and add tests.

flash_examples/finetuning/style_transfer.py
```diff
@@ -0,0 +1,25 @@
+import sys
+
+import flash
+from flash.data.utils import download_data
+from flash.utils.imports import _PYSTICHE_AVAILABLE
+from flash.vision.style_transfer import StyleTransfer, StyleTransferData
+
+if _PYSTICHE_AVAILABLE:
```
pmeier, line 8: If `pystiche` is not available, we'll never get to this, since this PR makes `flash` dependent on `pystiche` at import. Should this not be the case?

flash/vision/style_transfer/model.py
edgarriba: yes

flash_examples/finetuning/style_transfer.py
tchaton: We will refactor requirement, but Flash will depend on  pystiche only if user do `pip install flash['style-transfer']` and we need to make sure the code doesn't break your library isn't installed.

flash/vision/style_transfer/data.py
```diff
@@ -0,0 +1,102 @@
...
+from flash.data.process import Preprocess
+from flash.data.transforms import ApplyToKeys
+from flash.vision.classification import ImageClassificationData, ImageClassificationPreprocess
+
+from ._utils import raise_not_supported
+
+__all__ = ["StyleTransferPreprocess", "StyleTransferData"]
+
+
+def _apply_to_input(default_transforms_fn) -> Callable[..., Dict[str, ApplyToKeys]]:
```
tchaton, line 19: Maybe worth to move this to transform.py as it could be reused.

flash/vision/style_transfer/model.py
```diff
@@ -0,0 +1,227 @@
...
+        serializer: Optional[Union[Serializer, Mapping[str, Serializer]]] = None,
+    ):
+        self.save_hyperparameters(ignore="style_image")
+
+        if isinstance(style_image, str):
+            style_image = read_image(style_image)
+
+        if model is None:
+            # TODO: import this from pystiche
+            model = Transformer()
```
tchaton, line 146: `pystiche` should a hub of model (pretrained possibly) and make them available through the registry.

flash/vision/style_transfer/model.py
```diff
@@ -0,0 +1,227 @@
...
+
+        if multi_layer_encoder is None:
+            multi_layer_encoder = self.default_multi_layer_encoder()
```
tchaton, line 149: I guess the default could also come with the registry.

flash/vision/style_transfer/model.py
```diff
@@ -0,0 +1,227 @@
...
+            optimizer=optimizer,
+            optimizer_kwargs=optimizer_kwargs,
+            scheduler=scheduler,
+            scheduler_kwargs=scheduler_kwargs,
+            learning_rate=learning_rate,
+            serializer=serializer,
+        )
+
+        self.perceptual_loss = perceptual_loss
+        self.perceptual_loss.set_style_image(style_image)
```
tchaton, line 171: Is it possible to set multiple style_images and to randomly sample from them ?

flash/vision/style_transfer/model.py
```diff
@@ -0,0 +1,227 @@
...
+        content_layer = "relu2_2"
+        content_encoder = multi_layer_encoder.extract_encoder(content_layer)
+        content_weight = 1e5
+        return ops.FeatureReconstructionOperator(content_encoder, score_weight=content_weight)
+
+    def default_style_loss(
+        self, multi_layer_encoder: Optional[enc.MultiLayerEncoder] = None
+    ) -> ops.MultiLayerEncodingOperator:
+
+        class GramOperator(ops.GramOperator):
```
tchaton, line 190: Should come from `pystiche`

flash/vision/style_transfer/model.py
```diff
@@ -0,0 +1,227 @@
...
+                repr = super().enc_to_repr(enc)
+                num_channels = repr.size()[1]
+                return repr / num_channels
+
+        if multi_layer_encoder is None:
+            multi_layer_encoder = self.default_multi_layer_encoder()
+
+        style_layers = ("relu1_2", "relu2_2", "relu3_3", "relu4_3")
+        style_weight = 1e10
+        return ops.MultiLayerEncodingOperator(
```
tchaton, line 202: Worth to expose those elements to the task as `style_loss_kwargs`

flash/vision/style_transfer/model.py
```diff
@@ -0,0 +1,227 @@
...
+        )
+
+    def step(self, batch: Any, batch_idx: int) -> Any:
+        input_image = batch
+        self.perceptual_loss.set_content_image(input_image)
+
+        output_image = self(batch)
+        loss = self.perceptual_loss(output_image).total()
+
+        logs = dict(perceptual_loss=loss)
```
tchaton, line 217: Do you have any metrics ? Also better to use self.log to log metrics. Returning a dictionary is -/+ depreciated.

flash/vision/style_transfer/model.py
```diff
@@ -0,0 +1,227 @@
...
+        return dict(loss=loss, logs=logs)
+
+    def training_step(self, batch: Any, batch_idx: int) -> Any:
+        return super().training_step(batch[DefaultDataKeys.INPUT], batch_idx)
+
+    def validation_step(self, batch: Any, batch_idx: int) -> NoReturn:
+        raise_not_supported("validation")
```
tchaton, line 224: Not sure entirely why validation isn't supported. 
Also, might be better to set `self.validation_step = None` in the init function.

flash_examples/finetuning/style_transfer.py
```diff
@@ -0,0 +1,25 @@
...
+    import pystiche.demo
+else:
+    print("Please, run `pip install pystiche`")
+    sys.exit(0)
+
+download_data("http://images.cocodataset.org/zips/train2014.zip", "data")
+
+data_module = StyleTransferData.from_folders(train_folder="data", batch_size=4)
+
+style_image = pystiche.demo.images()["paint"].read(size=256)
```
tchaton, line 18: Can we replace this by an image from internet, like 

.

style_image could actually be url to style_transfer.

tchaton (commented):
Lot of great progress. Getting there.

flash/vision/style_transfer/model.py
tchaton: Same for all configurable elements.

flash/vision/style_transfer/model.py
pmeier: There are approaches like that, but they go beyond a simple entry example.

flash/vision/style_transfer/model.py
pmeier: Can't. I'm going for compatibility here with the PyTorch example / paper implementation and they both do this. This is bug / oversight in their implementation.

flash/vision/style_transfer/model.py
pmeier: > Do you have any metrics ?

There are none.

>  Also better to use self.log to log metrics. Returning a dictionary is -/+ depreciated.

Really? I mean I basically just did a simpler version of `super().step()` here, which also returns a dictionary.

flash/vision/style_transfer/model.py
pmeier: Style transfer is unsupervised so there is no reason for a validation. What labels what you validate against?

flash_examples/finetuning/style_transfer.py
pmeier: Sure, just tell me which image you want to use. Not sure about licensing though. The images I used are licensed very permissive.

flash/vision/style_transfer/model.py
pmeier: Yes, it should, but it probably won't in the next time. I only have limited time / resources to maintain `pystiche` so it is only one model for now.

flash/vision/style_transfer/model.py
pmeier: How is that easier than defining the loss outside of the task and pass it directly?

flash/vision/style_transfer/model.py
pmeier: Not sure what you mean. Could you elaborate?

flash_examples/finetuning/style_transfer.py
pmeier: But until this happens, this check is moot. Even after the refactor it is probably useless, since importing `StyleTransfer` depends on `pystiche`, so we need to do the guarding there.

flash_examples/finetuning/style_transfer.py
edgarriba: as soon we reference well should be fine. Consider also adding a link to here in the docs: https://paperswithcode.com/task/style-transfer. Papers with code will be reference soon also to submit results.

pmeier: > Consider also adding a link to here in the docs: https://paperswithcode.com/task/style-transfer. Papers with code will be reference soon also to submit results.

The link is doing image-optimization whereas we are implementing model-optimization here.

flash/vision/style_transfer/data.py
```diff
@@ -0,0 +1,102 @@
...
+            to_tensor_transform=torchvision.transforms.ToTensor(),
+            per_sample_transform_on_device=nn.Sequential(
+                transforms.Resize(min(self.image_size)),
+                transforms.CenterCrop(self.image_size),
+            ),
+        )
+
+    @property
+    def default_val_transforms(self) -> None:
+        # Style transfer doesn't support a validation phase, so we return nothing here
```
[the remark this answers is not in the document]
pmeier: IMO it will always be like this. The problem is that in order to validate or test, you need an objective metric to measure the model performance. With style transfer this metric needs to answer the question "how well is this image stylized given this were the original content and style". This is impossible to do in the current state of the art and I wager a guess it will stay that way.

flash/vision/style_transfer/model.py
```diff
@@ -0,0 +1,227 @@
...
+
+
+class StyleTransfer(Task):
+
+    def __init__(
+        self,
+        style_image: Union[str, torch.Tensor],
+        model: Optional[nn.Module] = None,
+        multi_layer_encoder: Optional[enc.MultiLayerEncoder] = None,
+        content_loss: Optional[Union[ops.ComparisonOperator, ops.OperatorContainer]] = None,
```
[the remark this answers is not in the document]
pmeier: This goes against what @tchaton proposed in https://github.com/PyTorchLightning/lightning-flash/pull/262#issuecomment-833300669. I'm happy to do either, but you (plural) need to decide which variant you would like to have.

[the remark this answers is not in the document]
edgarriba: ops, missed that detail. What do you think @tchaton  ? My proposal goes in the direction to have a more generic api across tasks..

flash/vision/style_transfer/data.py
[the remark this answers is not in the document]
edgarriba: sounds good - would be great to have that piece of information somewhere in the code or docs, or in both :)

flash/vision/style_transfer/model.py
pmeier: I've added an explanation through a comment.

flash_examples/finetuning/style_transfer.py
pmeier: @tchaton I need a reference image to check if the training works well with the default parameters.

tchaton: Oh I see. Makes sense.

flash/vision/style_transfer/backbone.py
```diff
@@ -0,0 +1,23 @@
...
+
+from flash.core.registry import FlashRegistry
+
+__all__ = ["STYLE_TRANSFER_BACKBONES"]
+
+MLE_FN_PATTERN = re.compile(r"^(?P<name>\w+?)_multi_layer_encoder$")
+
+STYLE_TRANSFER_BACKBONES = FlashRegistry("backbones")
+
+for mle_fn in dir(enc):
```
tchaton, line 13: Neat !

flash/vision/style_transfer/data.py
```diff
@@ -0,0 +1,111 @@
...
+            train_transform=train_transform,
+            val_transform=val_transform,
+            test_transform=test_transform,
+            predict_transform=predict_transform,
+            image_size=image_size,
+        )
+
+    @property
+    @_apply_to_input
+    def default_train_transforms(self) -> Dict[str, Callable]:
```
tchaton, line 56: @ethanwharris @pmeier We made a new addition. This code can be simplified.

Now, {}_default_transforms is a Preprocess hook.

You can also do:
```
def default_transforms(self):
    if not self.training or self.predicting: return None 
    if self.training:
        dict(
            to_tensor_transform=torchvision.transforms.ToTensor(),
            per_sample_transform_on_device=nn.Sequential(
                transforms.Resize(min(self.image_size)),
                transforms.CenterCrop(self.image_size),
            ),
        )
    else:
         dict(
            to_tensor_transform=torchvision.transforms.ToTensor(),
            per_sample_transform_on_device=transforms.Resize(min(self.image_size)),
        )

```

flash/vision/style_transfer/model.py
```diff
@@ -0,0 +1,122 @@
...
+
+    def __init__(
+        self,
+        style_image: Optional[Union[str, torch.Tensor]] = None,
+        model: Optional[nn.Module] = None,
+        backbone: str = "vgg16",
+        content_layer: str = "relu2_2",
+        content_weight: float = 1e5,
+        style_layers: Sequence[str] = ("relu1_2", "relu2_2", "relu3_3", "relu4_3"),
+        style_weight: float = 1e10,
```
tchaton, line 32: Looks really neat !

flash/vision/style_transfer/model.py
tchaton: Make sense :)

flash/core/utilities/imports.py
```diff
@@ -74,6 +74,7 @@ def _compare_version(package: str, op, version) -> bool:
 _PYTORCHVIDEO_AVAILABLE = _module_available("pytorchvideo")
 _MATPLOTLIB_AVAILABLE = _module_available("matplotlib")
 _TRANSFORMERS_AVAILABLE = _module_available("transformers")
+_PYSTICHE_AVAILABLE = _module_available("pystiche")
```
pmeier, line 77: Should this also be added to `_IMAGE_AVAILABLE`?

flash_examples/finetuning/style_transfer.py
pmeier: After #297 `pystiche` is now an optional dependency.

flash/vision/style_transfer/data.py
pmeier: I've encountered a problem with this: since I'm inheriting `ImageClassificationPreprocess` the `default_transforms_train` takes a higher priority. I think this is not intended. It happens because "prefix" (should be postfix BTW) is an empty string

https://github.com/PyTorchLightning/lightning-flash/blob/8eef52fa9b72befab339e388ce862e09f5f28e89/flash/core/data/data_pipeline.py#L183

but the check if for overridden methods checks for `None` explicitly:

https://github.com/PyTorchLightning/lightning-flash/blob/8eef52fa9b72befab339e388ce862e09f5f28e89/flash/core/data/data_pipeline.py#L131

Thus, `current_method_name` becomes `_default_transforms` which is indeed not overridden.

pmeier: Fixed in 21d0817.

flash/core/data/data_pipeline.py
```diff
@@ -180,7 +180,7 @@ def _resolve_function_hierarchy(
         if object_type is None:
             object_type = Preprocess
 
-        prefixes = ['']
```
pmeier, line 183: See https://github.com/PyTorchLightning/lightning-flash/pull/262#discussion_r633089549.

pmeier:
@tchaton @ethanwharris @edgarriba I need some input:

- Is the documentation skeleton that I added sufficient?
- What exactly should be tested? I've added a disabled finetuning testing, since it will train for two epochs otherwise.

requirements/datatype_image.txt
```diff
@@ -5,3 +5,5 @@ Pillow>=7.2
 kornia>=0.5.1
 matplotlib
 pycocotools>=2.0.2 ; python_version >= "3.7"
+git+https://github.com/pystiche/pystiche@releases/v0.7.2
+# pystiche>=0.7.2
```
pmeier, line 9: I'll push a new release as soon as we are sure nothing needs to be fixed / added to `pystiche` anymore.

docs/source/reference/style_transfer.rst
```diff
@@ -0,0 +1,35 @@
+##############
+Style Transfer
+##############
+
+********
+The task
+********
+
+TODO
+
```
tchaton, line 10: Mind describing the task ?

docs/source/reference/style_transfer.rst
```diff
@@ -0,0 +1,35 @@
...
+------
+
+***
+Fit
+***
+
+TODO
```
tchaton, line 17: Same here ?

flash/core/utilities/imports.py
tchaton: I think not doing so for now is better. I need to think about a way to segregate some deps

flash/image/style_transfer/data.py
```diff
@@ -0,0 +1,103 @@
...
+import torchvision
+from torch import nn
+from torchvision import transforms
+
+from flash.core.data.data_source import DefaultDataKeys, DefaultDataSources
+from flash.core.data.process import Preprocess
+from flash.core.data.transforms import ApplyToKeys
+from flash.image.classification import ImageClassificationData, ImageClassificationPreprocess
+
+from ._utils import raise_not_supported
```
tchaton, line 14: Use full path

tests/examples/test_scripts.py
```diff
@@ -94,6 +94,11 @@ def run_test(filepath):
             "translation.py",
             marks=pytest.mark.skipif(not _TEXT_AVAILABLE, reason="text libraries aren't installed")
         ),
+        # pytest.param(
```
tchaton, line 97: Could you add some tests ?

tchaton: By default, it runs fast_dev_run=True. I guess couple of tests to get the coverage up.
Plus, need to make sure the imports don t break if your lib is missing

flash/core/data/data_pipeline.py
```diff
@@ -190,9 +191,11 @@ def _resolve_function_hierarchy(
         elif stage == RunningStage.PREDICTING:
             prefixes += ['predict']
 
+        prefixes += [None]
```
pmeier, line 194: This would mean that `default_transforms` has a lower priority as `default_transforms_train`. IMO this will not fix https://github.com/PyTorchLightning/lightning-flash/pull/262#discussion_r633089549. If I'm not mistaken this will pull the default train transform which is unwanted

flash/image/style_transfer/backbone.py
```diff
@@ -0,0 +1,24 @@
...
+if _PYSTICHE_AVAILABLE:
+
+    from pystiche import enc
+
+    for mle_fn in dir(enc):
+
+        if "multi_layer_encoder" not in mle_fn:
+            continue
+
+        name = mle_fn.split("_")[0]
```
pmeier, line 17: This will result in the same name for `vgg19` and `vgg19_bn`. Why not keep the regex matching?

flash/image/style_transfer/backbone.py
```diff
@@ -0,0 +1,24 @@
...
+
+        STYLE_TRANSFER_BACKBONES(
+            fn=lambda: (getattr(enc, mle_fn)(), None),
+            name=mle_fn.split("_")[0],
```
pmeier, line 21:
```suggestion
            name=name,
```

flash_examples/predict/semantic_segmentation.py
```diff
@@ -24,7 +24,9 @@
 )
 
 # 2. Load the model from a checkpoint
-model = SemanticSegmentation.load_from_checkpoint("semantic_segmentation_model.pt")
+model = SemanticSegmentation.load_from_checkpoint(
```
pmeier, line 27: Not sure what this has to do with the PR?

requirements/datatype_image_style_transfer.txt
```diff
@@ -0,0 +1,4 @@
+torchvision
+Pillow>=7.2
+kornia>=0.5.1
+pystiche
```
pmeier, line 4:
```suggestion
pystiche>=0.7.2
```

tests/examples/test_scripts.py
```diff
@@ -94,6 +97,11 @@ def run_test(filepath):
             "translation.py",
             marks=pytest.mark.skipif(not _TEXT_AVAILABLE, reason="text libraries aren't installed")
         ),
+        pytest.param(
+            "finetuning",
+            "style_transfer.py",
+            marks=pytest.mark.skipif(not _IMAGE_STLYE_TRANSFER, reason="pystiche is not installed")
+        ),
```
pmeier, line 104: So this should run with every PR run? Each run will take ~2 hours.

pmeier (commented):
Hey @tchaton, thanks for the commits. I have some comments below. Additionally, it looks like you have added a lot of changes that seemingly have nothing to do with this PR. Was that intentional?

flash/image/style_transfer/backbone.py
tchaton: Reverted back.

flash/image/style_transfer/backbone.py
tchaton: Reverted back.

flash_examples/predict/semantic_segmentation.py
tchaton: It has a tiny bug, nothing major

requirements/datatype_image_style_transfer.txt
tchaton: Done !

tests/examples/test_scripts.py
tchaton: It is running with `FLASH_TESTING=1` which maps to `fast_dev_run=True` within the trainer.

flash/core/data/data_pipeline.py
tchaton: Removed ImageClassificationPrepreprocess, so we don't have this issue.

tchaton: Yes, `train_defaut_transforms` should have a higher priority than `default_transforms`.

docs/source/reference/style_transfer.rst
edgarriba: link to here would be nice too: https://paperswithcode.com/task/style-transfer

flash/image/style_transfer/model.py
```diff
@@ -0,0 +1,174 @@
...
+from flash.core.utilities.imports import _IMAGE_STLYE_TRANSFER
+from flash.image.style_transfer import STYLE_TRANSFER_BACKBONES
+
+if _IMAGE_STLYE_TRANSFER:
+    import pystiche.demo
+    from pystiche import enc, loss, ops
+    from pystiche.image import read_image
+else:
+
+    class enc:
```
edgarriba, line 20: `CamelCase` class name

flash/image/style_transfer/model.py
```diff
@@ -0,0 +1,174 @@
...
+    def __init__(
+        self,
+        style_image: Optional[Union[str, torch.Tensor]] = None,
+        model: Optional[nn.Module] = None,
+        backbone: str = "vgg16",
+        content_layer: str = "relu2_2",
+        content_weight: float = 1e5,
+        style_layers: Union[Sequence[str], str] = ("relu1_2", "relu2_2", "relu3_3", "relu4_3"),
+        style_weight: float = 1e10,
+        optimizer: Union[Type[torch.optim.Optimizer], torch.optim.Optimizer] = torch.optim.Adam,
```
edgarriba, line 75: in same tasks we are promoting `AdamW` since it has a better implementation of the original algorithm. Have you ever tried here ?

edgarriba (approved):
LGTM

flash/image/style_transfer/model.py
tchaton: Following pystiche.

ethanwharris (approved):
Awesome! Small comment

flash/image/style_transfer/model.py
pmeier: This just happens because these classes are actually modules.

docs/source/reference/style_transfer.rst
pmeier: @edgarriba see https://github.com/PyTorchLightning/lightning-flash/pull/262#discussion_r631697099

flash/image/style_transfer/model.py
pmeier: Nope. Until @tchaton changes this followed the official PyTorch example. Since we are deviating now, we could also use another optimizer.

PR #351, opened 2021-06-01
enforce random sampling at first call in kornia parallell transforms

edgarriba:
## What does this PR do?

enables to reuse the transforms parameters for kornia transforms in different calls

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/flash/core/data/transforms.py
+++ b/flash/core/data/transforms.py
@@ -79,7 +79,7 @@ class KorniaParallelTransforms(nn.Sequential):
         for transform in self.children():
             inputs = result
 
-            # we enforce the firs time to sample random params
+            # we enforce the first time to sample random params
             result[0] = transform(inputs[0])
 
             if hasattr(transform, "_params") and bool(transform._params):
--- a/flash/core/data/transforms.py
+++ b/flash/core/data/transforms.py
@@ -73,30 +73,27 @@ class KorniaParallelTransforms(nn.Sequential):
 
     def __init__(self, *args):
         super().__init__(*[convert_to_modules(arg) for arg in args])
-        self._reuse_params: bool = False
-
-    @property
-    def reuse_params(self) -> bool:
-        return self._reuse_params
-
-    def set_reuse_params(self, val: bool):
-        self._reuse_params = val
 
     def forward(self, inputs: Any):
         result = list(inputs) if isinstance(inputs, Sequence) else [inputs]
         for transform in self.children():
             inputs = result
+
+            # we enforce the firs time to sample random params
+            result[0] = transform(inputs[0])
+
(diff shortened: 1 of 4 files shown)


flash/core/data/transforms.py
```diff
@@ -73,6 +73,14 @@ class KorniaParallelTransforms(nn.Sequential):
 
     def __init__(self, *args):
         super().__init__(*[convert_to_modules(arg) for arg in args])
+        self._reuse_params: bool = False
+
+    @property
+    def reuse_params(self) -> bool:
+        return self._reuse_params
+
+    def set_reuse_params(self, val: bool):
```
ethanwharris, line 82: Why not use `@reuse_params.setter` here?

ethanwharris (commented):
Looks good, some failing tests. Is there any reason anyonme would want to do this though? It looks like `reuse_params=True` will make it apply the same transform every time, so you're not really getting an augmentation just a constant transform.

flash/core/data/transforms.py
edgarriba: if we don't have this e.g in segmentation I couldn't find a way to keep parameters to be applied to both images and masks. ok, for setter

edgarriba: for the failing tests no idea - cannot reproduce in local

ethanwharris: Looks like you need to update the segmentation model on AWS. But the usage here applies the same transform to image and mask no? https://github.com/PyTorchLightning/lightning-flash/blob/c11498d0c12357f448dedea0fa61262714f4fa11/flash/image/segmentation/transforms.py#L55

ethanwharris: We definitely don't want to just always apply the same transform though

edgarriba: ok, I was assuming for the case to call multiple times `KorniaParallelTransforms`

ethanwharris (approved):
LGTM :smiley:

flash/image/segmentation/transforms.py
```diff
@@ -38,7 +38,7 @@ def default_transforms(image_size: Tuple[int, int]) -> Dict[str, Callable]:
         "post_tensor_transform": nn.Sequential(
             ApplyToKeys(
                 [DefaultDataKeys.INPUT, DefaultDataKeys.TARGET],
-                KorniaParallelTransforms(K.geometry.Resize(image_size, interpolation='bilinear')),
+                KorniaParallelTransforms(K.geometry.Resize(image_size, interpolation='nearest')),
```
edgarriba, line 41: related to https://github.com/PyTorchLightning/lightning-flash/pull/352

PR #436, opened 2021-06-21
Fix pip install error messaging

### LINKED ISSUE #435: Segmentation recommends Kornia install rather than image
Outside of the scope of this issue but taking the first example from the doc:
https://lightning-flash.readthedocs.io/en/latest/reference/flash_to_production.html#
```
from flash.image import SemanticSegmentation
from flash.image.segmentation.serialization import SegmentationLabels

model = SemanticSegmentation.load_from_checkpoint(
    "https://flash-weights.s3.amazonaws.com/semantic_segmentation_model.pt"
)
```
It is asking for kornia as a dependency.
Should we install it by default or suggest a command how to install it (the same way you do for serve?)
Screen Shot 2021-06-18 at 3 40 48 PM

Note that once I did install kornia it asked me for `pip install 'lightning-flash[image]'` but only after.

_Originally posted by @zippeurfou in https://github.com/PyTorchLightning/lightning-flash/issues/428#issuecomment-864239447_

ethanwharris:
## What does this PR do?

Fixes #435 

## Before submitting

## PR review

## Did you have fun?


.github/workflows/ci-testing.yml
```diff
@@ -37,10 +37,6 @@ jobs:
             python-version: 3.8
             requires: 'latest'
             topic: 'text'
-          - os: ubuntu-20.04
-            python-version: 3.8
-            requires: 'latest'
-            topic: 'image_style_transfer'
```
Borda, line 43: any reason?

flash/core/utilities/imports.py
```diff
@@ -99,5 +98,31 @@ def _compare_version(package: str, op, version) -> bool:
...
 ])
 _SERVE_AVAILABLE = _FASTAPI_AVAILABLE and _PYDANTIC_AVAILABLE and _CYTOOLZ_AVAILABLE and _UVICORN_AVAILABLE
+
+_EXTRAS_AVAILABLE = {
+    'image': _IMAGE_AVAILABLE,
+    'tabular': _TABULAR_AVAILABLE,
+    'text': _TEXT_AVAILABLE,
+    'video': _VIDEO_AVAILABLE,
+    'serve': _SERVE_AVAILABLE,
+}
```
Borda, line 111: enum?

flash/core/utilities/imports.py
```diff
@@ -99,5 +98,31 @@ def _compare_version(package: str, op, version) -> bool:
...
+
+
+def _requires_extras(extras: str):
```
Borda, line 114: who about multiple? double wrapper or just allow `Union[str, Sequence[str]]`

flash/core/utilities/imports.py
```diff
@@ -85,9 +86,7 @@ def _compare_version(package: str, op, version) -> bool:
 
 if Version:
     _TORCHVISION_GREATER_EQUAL_0_9 = _compare_version("torchvision", operator.ge, "0.9.0")
-    _PYSTICHE_GREATER_EQUAL_0_7_2 = _compare_version("pystiche", operator.ge, "0.7.2")
 
-_IMAGE_STLYE_TRANSFER = _PYSTICHE_AVAILABLE
 _TEXT_AVAILABLE = _TRANSFORMERS_AVAILABLE
 _TABULAR_AVAILABLE = _TABNET_AVAILABLE and _PANDAS_AVAILABLE
```
Borda, line 91: this makes it confusing, how about if you have installed only one... then anyone using the wrapper shall carefully set that you really need both...

flash/core/serve/core.py
```diff
@@ -100,15 +100,13 @@ class GridModel:
     *  How to handle ``__init__`` args not recorded in hparams of ``pl.LightningModule``
     """
 
+    @_requires_extras("serve")
```
Borda, line 103: do you really need all, even COCO?

flash/image/classification/data.py
```diff
@@ -16,14 +16,13 @@
...
 from pytorch_lightning.trainer.states import RunningStage
-from pytorch_lightning.utilities.exceptions import MisconfigurationException
 
 from flash.core.data.base_viz import BaseVisualization  # for viz
 from flash.core.data.callback import BaseDataFetcher
 from flash.core.data.data_module import DataModule
 from flash.core.data.data_source import DefaultDataKeys, DefaultDataSources
 from flash.core.data.process import Deserializer, Preprocess
-from flash.core.utilities.imports import _MATPLOTLIB_AVAILABLE, _PIL_AVAILABLE
+from flash.core.utilities.imports import _MATPLOTLIB_AVAILABLE, _PIL_AVAILABLE, _requires_extras
```
Borda, line 25:
```suggestion
from flash.core.utilities.imports import _PIL_AVAILABLE, _requires_extras
```
probably... but flake8 shall tell you

flash/image/classification/data.py
```diff
@@ -121,14 +121,12 @@ def _to_numpy(img: Union[torch.Tensor, Image.Image]) -> np.ndarray:
             raise TypeError(f"Unknown image type. Got: {type(img)}.")
         return out
 
+    @_requires_extras("image")
     def _show_images_and_labels(self, data: List[Any], num_samples: int, title: str):
```
Borda, line 125: do you need kornia here?

.github/workflows/ci-testing.yml
ethanwharris: I've removed the "image_style_transfer" extras, they are now in "image" so we don't need seperate CI anymore

flash/core/serve/core.py
ethanwharris: Not sure what you mean here, this function just requires serve. Serve tests import "all" extras so that we can test all servable tasks (which each need their own extras)

flash/core/utilities/imports.py
ethanwharris: If just one of the tabular dependencies is installed and you try to use the TabularClassifier then it will error saying it needs all tabular dependencies to run. I don't think there are parts of the tabular classifier that you can use without installing those dependencies.

flash/core/utilities/imports.py
ethanwharris: Currently we just support requiring a single "extras" install. It would only make sens to add support for multiple when / if we have a use case for it.

Borda: basically I aim at the granularity that the user may have installed some packages without Flash defined extras so for example for this function he does need all extras for image, right?

flash/core/utilities/imports.py
Borda: so you say that ll tabular always need both dependencies, right?

flash/image/classification/data.py
Borda: basically, I aim at the granularity that the user may have installed some packages without Flash defined extras so for example for this function he does need all extras for image, right?

ethanwharris: The problem with granularity is that you get this sort of flow:
- try to do something e.g. create `ImageClassifier`
- get error "install torchvision"
- now create `ImageClassifier` and try to train it
- get error "install kornia"
- now create and train and try to visualize
- get error "install matplotlib"

So with this PR, instead you should just get told to install the package with "image" extras at the beginning.

flash/core/utilities/imports.py
ethanwharris: Yeah, if you want to train a model then it will need both tabnet and pandas

flash/image/classification/data.py
Borda: I see so maybe we can have another wrapper just for one or more particular packages?

Borda (approved):
nice simplification :rabbit:

flash/core/utilities/imports.py
justusschock: Wouldn't do it here since a mapping is perfectly fine when we map strings to boolean

justusschock (approved):

Issue #361, opened 2021-06-04
Better documentation for metrics argument


ethanwharris:
## 📚 Documentation

We should provide better documentation for using custom metrics (via the `metrics` argument) with a task. See related issue #358

PR #473, opened 2021-06-27
Better description of the metric arg

SkafteNicki:
## What does this PR do?

Fixes https://github.com/PyTorchLightning/lightning-flash/issues/361
Adds better description of the metric arg for all tasks. Additionally, to me it seems that for the following task the metric arg has no effect (because they do not use the `step` method from the base `Task` class)
* Object detection
* seq2seq summarization
* seq2seq translation

have added this to the docs but need confirmation from someone else that this is actually true.

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/flash/image/classification/model.py
+++ b/flash/image/classification/model.py
@@ -15,9 +15,9 @@ from types import FunctionType
 from typing import Any, Callable, Dict, List, Mapping, Optional, Sequence, Tuple, Type, Union
 
 import torch
-import torchmetrics
 from torch import nn
 from torch.optim.lr_scheduler import _LRScheduler
+from torchmetrics import Metric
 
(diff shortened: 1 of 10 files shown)


ethanwharris (approved):
This is great, thanks @SkafteNicki :smiley: what do you suggest about the unused args? I think for now it would be good to remove them, do you want to do that in another PR?

SkafteNicki:
@ethanwharris it depends, if we think that in the future these tasks should have a working metrics argument then it is maybe better to just leave them as they are, instead of going through the process of deprecating them :]
If not, then lets just remove them.

ethanwharris:
@SkafteNicki Yeah, that makes sense, I've opened an issue (#480) to get them working with metrics.

PR #491, opened 2021-06-28
Add lazy importing for FiftyOne

ehofesmann:
## What does this PR do?

This PR adds the ability to lazily import modules so that they are not actually imported until their first reference. This is applied to all `fiftyone` imports to avoid spinning up the FiftyOne database service if it is not being used.

## Before submitting
- [X] Did you read the [contributor guideline](https://github.com/PyTorchLightning/lightning-flash/tree/master/.github/CONTRIBUTING.md), Pull Request section?
- [X] Did you make sure your PR does only one thing, instead of bundling different changes together?
- [X] Did you make sure to update the documentation with your changes?
- [X] Did you verify new and existing tests pass locally with your changes?

## PR review

## Did you have fun?

### DIFF
--- a/flash/core/classification.py
+++ b/flash/core/classification.py
@@ -21,12 +21,12 @@ from pytorch_lightning.utilities import rank_zero_warn
 from flash.core.data.data_source import DefaultDataKeys, LabelsState
 from flash.core.data.process import Serializer
 from flash.core.model import Task
-from flash.core.utilities.imports import _FIFTYONE_AVAILABLE
+from flash.core.utilities.imports import _FIFTYONE_AVAILABLE, lazy_import
 
 if _FIFTYONE_AVAILABLE:
-    from fiftyone.core.labels import Classification, Classifications
+    fol = lazy_import("fiftyone.core.labels")
 else:
-    Classification, Classifications = None, None
+    fol = None
 
 
 def binary_cross_entropy_with_logits(x: torch.Tensor, y: torch.Tensor) -> torch.Tensor:
@@ -219,7 +219,7 @@ class FiftyOneLabels(ClassificationSerializer):
     def serialize(
         self,
         sample: Any,
-    ) -> Union[Classification, Classifications, Dict[str, Any], Dict[str, Any]]:
+    ) -> Union[fol.Classification, fol.Classifications, Dict[str, Any], Dict[str, Any]]:
         pred = sample[DefaultDataKeys.PREDS] if isinstance(sample, Dict) else sample
         pred = torch.tensor(pred)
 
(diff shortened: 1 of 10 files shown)


flash/core/classification.py
```diff
@@ -21,12 +21,15 @@
...
 from flash.core.data.process import Serializer
 from flash.core.model import Task
-from flash.core.utilities.imports import _FIFTYONE_AVAILABLE
+from flash.core.utilities.imports import _FIFTYONE_AVAILABLE, lazy_import
 
+Classification, Classifications = None, None
 if _FIFTYONE_AVAILABLE:
-    from fiftyone.core.labels import Classification, Classifications
+    fol = lazy_import("fiftyone.core.labels")
+    if TYPE_CHECKING:
```
ethanwharris, line 29: This may cause some issues for sphinx docs build, but I'm not sure. Seems to be working fine for now.

ethanwharris (approved):
LGTM, not sure about use of `TYPE_CHECKING` I know it's caused issues with docs build in the past (I think sphinx doesn't resolve type hints that rely on imports from inside `if TYPE_CHECKING` clause.

flash/core/classification.py
ehofesmann: I didn't realize this. Did you figure out a way around it when it came up in the past?

ethanwharris: Damn, I thought I remembered something but that was for circular imports so doesn't apply here 😆 I _think_ (with low confidence) that if you annotate types as `'fol.Classification'` or `'fol.Classifications'` (with quotes) then it should avoid the import of fiftyone in normal usage (but will still be triggered when type checking). That _should_ allow you to remove the `TYPE_CHECKING` clause.

ehofesmann: Adding quotes does work to prevent the import, but then when I run it through `mypy` I get:
```
error: Name "foc.SampleCollection" is not defined
```

Is there another type checking tool that you use other than `mypy`?

ethanwharris: Ideally it would work with mypy, I guess just stick with `TYPE_CHECKING` for now and we can just fix it if we start getting issues.

ehofesmann: Sounds good, let me know if any issues pop up.

ethanwharris: Will do :smiley:

akihironitta:
@ehofesmann This can be used for other optional packages, too, right? There would be tons of optional external packages as Flash grows, so it could be applied to others than fiftyone maybe?

ehofesmann:
@akihironitta Definitely! It is a general-purpose function that can lazily import any module. As long as there are no issues with `TYPE_CHECKING` that come up, this should be able to prevent unnecessary package loading.

PR #583, opened 2021-07-13
Move big dependencies out to `*_extras` requirements and treat them as fully optional

### LINKED ISSUE #451: Installing [image] extras is not sufficient on Python 3.6?
## 🐛 Bug

On Python 3.6, `pip install lightning-flash[image]` is not sufficient to allow me to use an `ObjectDetector`:

```py
from flash.image import ObjectDetector

# error: tells me I need to `pip install 'lightning-flash[image]'`, but I already did!
model = ObjectDetector.load_from_checkpoint("https://flash-weights.s3.amazonaws.com/object_detection_model.pt")
```

I believe the issue is that pycocotools has a `python_version >= 3.7` marker in the [requirements file](https://github.com/PyTorchLightning/lightning-flash/blob/master/requirements/datatype_image.txt). Either that environment marker is necessary, in which case the image extras only support Python 3.7+, or the marker should be removed.

FWIW, I manually `pip install pycocotools` and was able to successfully use the checkpoint that I tried to load in the example code below.

### To Reproduce

Create a fresh environment:

```shell
conda create -n test python=3.6
conda activate test

pip install lightning-flash
pip install 'lightning-flash[image]'
```

Now try to use the image extras:

```py
from flash.image import ObjectDetector

# error: tells me I need to `pip install 'lightning-flash[image]'`, but I already did!
model = ObjectDetector.load_from_checkpoint("https://flash-weights.s3.amazonaws.com/object_detection_model.pt")
```

### Environment

 - OS (e.g., Linux): macOS 10.14.6
 - Python version: Python 3.6.13 :: Anaconda, Inc.
 - Lightning Flash version: 0.4.0

ethanwharris:
## What does this PR do?

Fixes #451

## Before submitting

## PR review

## Did you have fun?


flash/core/classification.py
```diff
@@ -195,6 +195,7 @@ class FiftyOneLabels(ClassificationSerializer):
             list of FiftyOne labels (False)
     """
 
+    @requires("fiftyone")
```
tchaton, line 198: Neat !

tchaton: quick question. How does this behave with `inspect.signature` ?

tchaton (approved):

flash/core/classification.py
ethanwharris: It should be fine as when the packages are installed the decorator is a no-op (just returns the original function). When not installed, the function will always error anyway but we use functools.wraps which preserves signatures :smiley:

PR #642, opened 2021-08-07
update pystiche dependency

pmeier:
## What does this PR do?

This upgrades a couple of things regarding the `pystiche` dependency:

- `pystiche==1.0.0` was released and now follows semantic versioning. Thus, It is reasonable to pin it to the current stable release.
- `pystiche.ops` is deprecated in favor of `pystiche.loss`. Although everything should work as expected as is, without change the code will now emit deprecation warnings. All occurrences of `pystiche.ops` were replaced.
- `pystiche`'s default branch was changed from `master` to `main`. A link was updated to reflect this.
- `pystiche` should be written with a lower `p`. All occurrences in the documentation that did not do this were updated.

Fixes # (issue)

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/README.md
+++ b/README.md
@@ -446,9 +446,9 @@ python flash_examples/finetuning/semantic_segmentation.py
 
 </details>
 
-### Example 7: Style Transfer with Pystiche
+### Example 7: Style Transfer with pystiche
 
-Flash has a [Style Transfer task](https://lightning-flash.readthedocs.io/en/latest/reference/style_transfer.html) for Neural Style Transfer (NST) with [Pystiche](https://github.com/pystiche/pystiche).
+Flash has a [Style Transfer task](https://lightning-flash.readthedocs.io/en/latest/reference/style_transfer.html) for Neural Style Transfer (NST) with [pystiche](https://pystiche.org).
 
 <details>
   <summary>View example</summary>
--- a/docs/source/reference/style_transfer.rst
+++ b/docs/source/reference/style_transfer.rst
@@ -12,7 +12,7 @@ The Task
 The Neural Style Transfer Task is an optimization method which extract the style from an image and apply it another image while preserving its content.
 The goal is that the output image looks like the content image, but “painted” in the style of the style reference image.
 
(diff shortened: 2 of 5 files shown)


pmeier:
Two problems here:

- Some workflows fail, because they can't import stuff from `pystiche>=1.0.0` although they never install any version explicitly. This tells me, that they pull in an older version from cache. I'm not in a position to fix that.
- JIT tests are failing, because parts of the `pystiche.loss.PerceptualLoss` are no longer `torch.jit.trace`'able. This worked before, but only by chance. I'm wondering, why do you test for losses being able to be traced? AFAIK, `torch.jit` is used for efficient deployment of models for which the loss that they were trained on is irrelevant.

ethanwharris (approved):
LGTM. Fixed the issues:
- It wasn't pystiche getting cached, the error was just cause by the use of pystiche components in the type annotation despite pystiche not being installed
- Regarding JIT, we don't try to jit the loss, only the forward. The issue seems to be that JIT.trace first recurses over all callable attributes of the model to see if they have been marked for export. When it hits loss.encoder, that throws a runtime error which is then not caught. I think the `Multiple enocder attributes` error in pystiche could be replaced with an AttributeError rather than a RuntimeError in order to fix this. For now I have just removed the loss_fn and perceptual_loss attributes before jitting in the test.

pmeier:
@ethanwharris 

Thanks for the input.

> It wasn't pystiche getting cached, the error was just cause by the use of pystiche components in the type annotation despite pystiche not being installed

Oh, I see. Wouldn't it make a lot more sense to quote the annotations instead of providing mock modules? That way there is no runtime effect in case `pystiche` is not available and if it is available `mypy` treats it the same as an unqouted annotation.

> Regarding JIT, we don't try to jit the loss, only the forward. The issue seems to be that JIT.trace first recurses over all callable attributes of the model to see if they have been marked for export. When it hits loss.encoder, that throws a runtime error which is then not caught. I think the Multiple enocder attributes error in pystiche could be replaced with an AttributeError rather than a RuntimeError in order to fix this. For now I have just removed the loss_fn and perceptual_loss attributes before jitting in the test.

That is a valid point. I'll go and fix this there and send a new patch here.

PR #670, opened 2021-08-17
updated mock Image object

### LINKED ISSUE #437: Update mock class for Image
## 🚀 Feature
At the moment the if the `_PIL_AVAILABEL` flag is false `Image` is replaced with a class:

```python
class Image: 
    Image = None
```
In the `_to_numpy` functions in the image part of the lib the type of the input is checked to select the appropriate action. This can lead to some side effects because if `PIL` is not installed the type is then checked against None, which might lead to some strange behaviour is the input is None for some reason. This might be hard to track down when it happens. I purpose to change the mock class to warn the user if the mock class/attribute is used (see below).

```python
class Image:
    
    @property
    def Image(self):
        warn("This is just a mock object. The function you called might not work properly. You are missing the PIL libary, you can install it with 'pip install Pillow'.")
```
I can create a PR for this.

### Motivation
Better feedback for the user in edge case.

### Pitch
Replace the mock class with the purposed one.

### Alternatives

An alternative could be to warn the user when the mock class is created. This would have the draw back that the user would need to track down where exactly the error might result from.

ananyahjha93:
## What does this PR do?

Fixes #437 

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/flash/core/utilities/imports.py
+++ b/flash/core/utilities/imports.py
@@ -97,6 +97,8 @@ _ROUGE_SCORE_AVAILABLE = _module_available("rouge_score")
 _SENTENCEPIECE_AVAILABLE = _module_available("sentencepiece")
 _DATASETS_AVAILABLE = _module_available("datasets")
 _ICEVISION_AVAILABLE = _module_available("icevision")
+_ICEDATA_AVAILABLE = _module_available("icedata")
+_TORCH_ORT_AVAILABLE = _module_available("torch_ort")
 
 if _PIL_AVAILABLE:
     from PIL import Image
@@ -138,6 +140,7 @@ _IMAGE_AVAILABLE = all(
         _PYSTICHE_AVAILABLE,
         _SEGMENTATION_MODELS_AVAILABLE,
         _ICEVISION_AVAILABLE,
+        _ICEDATA_AVAILABLE,
     ]
 )
 _SERVE_AVAILABLE = _FASTAPI_AVAILABLE and _PYDANTIC_AVAILABLE and _CYTOOLZ_AVAILABLE and _UVICORN_AVAILABLE
--- a/flash/core/utilities/imports.py
+++ b/flash/core/utilities/imports.py
@@ -15,9 +15,9 @@ import functools
 import importlib
 import operator
 import types
-from warnings import warn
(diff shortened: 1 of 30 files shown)


flash/image/classification/data.py
```diff
@@ -11,6 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+from logging import warn
```
Borda, line 14:
```suggestion
from warnings import warn
```

flash/image/segmentation/data.py
```diff
@@ -71,8 +72,19 @@
...
+    class MetaImage(type):
+        def __init__(cls, name, bases, dct):
+            super().__init__(name, bases, dct)
+
+            cls._Image = None
+
+        @property
+        def Image(cls):
+            warn("Mock object called due to missing PIL library. Install PIL using 'pip install Pillow'.")
+            return cls._Image
```
Borda, line 84: looks the same in all files, so why not have it only once?

ethanwharris: Agree, this could maybe go in `core/utilities/imports.py` that's where our other import warnings are. Can we also switch it to say to install `lightning-flash[image]` rather than PIL

flash/core/utilities/imports.py
```diff
@@ -99,6 +100,25 @@ def _compare_version(package: str, op, version) -> bool:
...
+
+    class MetaImage(type):
+        def __init__(cls, name, bases, dct):
+            super().__init__(name, bases, dct)
+
+            cls._Image = None
+
+        @property
+        def Image(cls):
+            warn("Mock object called due to missing PIL library. Please use \"pip install 'lightning-flash[image]'\".")
```
ethanwharris, line 115: Let's raise a `ModuleNotFoundError` here rather than warn. Safe to assume whatever code winds up using it won't work correctly.

ananyahjha93: can't do Module not found error, the property needs to return a None value, since Image.Image is being checked for its type in some part of the code, can't throw an error there.

ethanwharris (approved):

PR #737, opened 2021-09-06
[PoC] Add MetaLearning support through learn2learn

tchaton:
## What does this PR do?

Here is the PR from the learn2learn side: https://github.com/learnables/learn2learn/pull/257

## Before submitting

## PR review

## Did you have fun?


flash/core/utilities/providers.py
```diff
@@ -39,6 +39,7 @@ def __str__(self):
 _SEGMENTATION_MODELS = Provider(
     "qubvel/segmentation_models.pytorch", "https://github.com/qubvel/segmentation_models.pytorch"
 )
+_LEARN2LEARN = Provider("earnables/learn2learn", "https://github.com/learnables/learn2learn")
```
ethanwharris, line 42:
```suggestion
_LEARN2LEARN = Provider("learnables/learn2learn", "https://github.com/learnables/learn2learn")
```

flash/image/classification/adapters.py
```diff
@@ -0,0 +1,391 @@
...
+
+if _LEARN2LEARN_AVAILABLE:
+    from learn2learn import algorithms
+
+    for algorithm in dir(algorithms):
+        try:
+            if "lightning" in algorithm.lower() and issubclass(getattr(algorithms, algorithm), LightningModule):
+                TRAINING_STRATEGIES(
+                    name=algorithm.lower().replace("lightning", ""),
+                    fn=_fn,
```
ethanwharris, line 385: We could perhaps just do `fn=partial(Learn2LearnAdapter, algorithm=getattr(algorithms, algorithm))` here, then not need to get the metadata later on

ethanwharris (commented):
Awesome :smiley:

Borda (commented):
do we have a task, real example, competition to test it on?

flash/core/utilities/providers.py
tchaton: Good catch !

pietrolesci:
Adding to @Borda's comment, an easy way to test whether it's working properly is to replicate the examples in [this](https://github.com/learnables/learn2learn/tree/master/examples/vision) folder of the l2l library (note: link to vision examples).

ananyahjha93 (approved):

flash/core/data/process.py
```diff
@@ -344,14 +344,20 @@ def default_transforms() -> Optional[Dict[str, Callable]]:
 
     def pre_tensor_transform(self, sample: Any) -> Any:
         """Transforms to apply on a single object."""
+        if isinstance(sample, list):
+            return [self.current_transform(s) for s in sample]
         return self.current_transform(sample)
```
ethanwharris, line 349: Can we move this out somewhere since it's repeated in three places?

flash/image/classification/adapters.py
```diff
@@ -0,0 +1,416 @@
...
+from flash.core.model import Task
+from flash.core.registry import FlashRegistry
+from flash.core.utilities.imports import _LEARN2LEARN_AVAILABLE
+from flash.core.utilities.providers import _LEARN2LEARN
+from flash.core.utilities.url_error import catch_url_error
+
+if _LEARN2LEARN_AVAILABLE:
+    import learn2learn as l2l
+
+    class RemapLabels(l2l.data.transforms.TaskTransform):
```
ethanwharris, line 39: Could we always define this even if l2l is not available? That way it can be added to API reference

flash/image/classification/adapters.py
```diff
@@ -0,0 +1,416 @@
...
+
+class DefaultAdapter(Adapter):
+    """The ``DefaultAdapter`` is an :class:`~flash.core.adapter.Adapter`."""
+
+    required_extras: str = "image"
+
+    def __init__(self, task: AdapterTask, backbone: torch.nn.Module, head: torch.nn.Module):
+        super().__init__()
+
+        self._task = NoModule(task)
```
ethanwharris, line 353: Could we just move out the base `*_step` implementations to a mixin to just inherit the behaviour rather than have a reference to the task?

flash/image/classification/adapters.py
```diff
@@ -0,0 +1,416 @@
...
+    for algorithm in dir(algorithms):
+        # skip base class
+        if algorithm == "LightningEpisodicModule":
+            continue
+        try:
+            if "lightning" in algorithm.lower() and issubclass(getattr(algorithms, algorithm), LightningModule):
+                TRAINING_STRATEGIES(
+                    name=algorithm.lower().replace("lightning", ""),
+                    fn=partial(Learn2LearnAdapter.from_task, algorithm=getattr(algorithms, algorithm)),
+                    providers=[_LEARN2LEARN],
```
ethanwharris, line 413: Nice!

flash/image/classification/adapters.py
```diff
@@ -0,0 +1,416 @@
...
+
+    def __init__(
+        self,
+        task: AdapterTask,
+        backbone: torch.nn.Module,
+        head: torch.nn.Module,
+        algorithm_cls: Type[LightningModule],
+        ways: int,
+        kshots: int,
+        queries: int = 1,
```
ethanwharris, line 125: Could do with some docstrings here, I'm not sure what all these are haha

ethanwharris (commented):
Looking good :smiley:

flash/image/classification/adapters.py
```diff
@@ -0,0 +1,416 @@
...
+        dataset = l2l.data.MetaDataset(
+            dataset, indices_to_labels=indices_to_labels, labels_to_indices=labels_to_indices
+        )
+        taskset = l2l.data.TaskDataset(
+            dataset=dataset,
+            task_transforms=self._default_transform(dataset),
+            num_tasks=-1,
+            task_collate=self._identity_fn,
+        )
+        dataset = Epochifier(taskset, 100)
```
pietrolesci, line 205: ### `num_epochs`, `num_task`, `meta_batch_size`, `kshots` and `batch_size`
Here, it seems a bit arbitrary to force the number of tasks per epoch to be 100. We should aim to replicate the training loop define, for example, [here](https://github.com/learnables/learn2learn/blob/06893e847693a0227d5f35a6e065e6161bb08201/examples/vision/maml_omniglot.py#L84-L129). In other words, ideally, the user should be able to choose the
- [x] number of epochs -> done in the trainer
- [x] how many task to create in total -> done by specifying the `num_task` in the l2l backend, ideally to be exposed to user
- [ ] number of tasks per epoch aka `meta_batch_size` -> ideally, to be exposed to user and can default to `num_task`
- [x] how many observations per each task -> controlled by `kshots` and dealt with by l2l in the backend
- [ ] the batch_size within the tasks -> suppose that `kshots` for each task is big enough that data do not fit in memory. Note that this can potentially avoid the cumbersome fact that in the datamodule the user is forced to set `batch_size` to 1 otherwise `process_{train, val, test}_dataset` produces an AssertionError

pietrolesci (commented):
Neat, looking great!

flash/core/utilities/imports.py
```diff
@@ -98,6 +98,7 @@ def _compare_version(package: str, op, version) -> bool:
 _DATASETS_AVAILABLE = _module_available("datasets")
 _ICEVISION_AVAILABLE = _module_available("icevision")
 _ICEDATA_AVAILABLE = _module_available("icedata")
+_LEARN2LEARN_AVAILABLE = _module_available("learn2learn")
```
pietrolesci, line 101: Check the specific version of l2l as I had to update to latest in order to run locally

requirements/datatype_image_extras.txt
```diff
@@ -2,6 +2,7 @@ matplotlib
 fiftyone
 classy_vision
 vissl>=0.1.5
+git+https://github.com/learnables/learn2learn.git
```
pietrolesci, line 5: Use (the now published) official version of the library

flash/image/classification/adapters.py
pietrolesci: Hi @ethanwharris, as this is [something which is static](https://github.com/PyTorchLightning/lightning-flash/blob/b33beb7d507e470a6350351b4fd166ac28b88e88/flash/image/classification/adapters.py#L163-L169) (the user cannot change this) and you'll likely use it with l2l only, having it in the API refs might be confusing

flash/core/data/process.py
tchaton: Done !

flash/image/classification/adapters.py
tchaton: Done !

flash/image/classification/adapters.py
tchaton: Done !

flash/image/classification/adapters.py
tchaton: Not sure to understand you. I did it this way, so we can use `Task.training_step(self.task, batch, batch_idx)` in the training_step of the adapter.

flash/image/classification/adapters.py
```diff
@@ -0,0 +1,416 @@
...
+
+    required_extras: str = "image"
+
+    def __init__(
+        self,
+        task: AdapterTask,
+        backbone: torch.nn.Module,
+        head: torch.nn.Module,
+        algorithm_cls: Type[LightningModule],
+        ways: int,
```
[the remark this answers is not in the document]
tchaton: I actually added support for passing test_ways if needed.

flash/image/classification/adapters.py
```diff
@@ -0,0 +1,416 @@
...
+            )
+
+        # convert the dataset to MetaDataset
+        dataset = l2l.data.MetaDataset(
+            dataset, indices_to_labels=indices_to_labels, labels_to_indices=labels_to_indices
+        )
+        taskset = l2l.data.TaskDataset(
+            dataset=dataset,
+            task_transforms=self._default_transform(dataset),
+            num_tasks=-1,
```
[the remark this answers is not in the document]
tchaton: Good catch !

flash/image/classification/adapters.py
```diff
@@ -0,0 +1,416 @@
...
+        # this algorithm requires a special treatment
+        self._algorithm_has_validated = self.algorithm_cls != l2l.algorithms.LightningPrototypicalNetworks
+
+    def _default_transform(self, dataset) -> List[Callable]:
+        return [
+            l2l.data.transforms.FusedNWaysKShots(dataset, n=self.ways, k=self.kshots),
+            l2l.data.transforms.LoadData(dataset),
+            RemapLabels(dataset),
+            l2l.data.transforms.ConsecutiveLabels(dataset),
+        ]
```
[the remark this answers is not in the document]
tchaton: Add support for passing more.

flash/image/classification/adapters.py
```diff
@@ -0,0 +1,416 @@
...
+        metadata = getattr(dataset, "data", None)
+        if metadata is None or (metadata is not None and not isinstance(dataset.data, list)):
+            raise MisconfigurationException("Only dataset built out of metadata is supported.")
+
+        indices_to_labels = {index: sample[DefaultDataKeys.TARGET] for index, sample in enumerate(dataset.data)}
+        labels_to_indices = defaultdict(list)
+        for idx, label in indices_to_labels.items():
+            labels_to_indices[label].append(idx)
+
+        if len(labels_to_indices) < self.ways:
```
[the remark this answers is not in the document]
tchaton: I would prefer ways to not be a dict.

flash/image/classification/adapters.py
tchaton: Done ! Thanks.

flash/core/utilities/imports.py
tchaton: Done !

requirements/datatype_image_extras.txt
tchaton: I need my PR to be merged in their repo first.

flash/image/classification/adapters.py
ethanwharris: Yeah, but we could just move the Task.*_step methods out somehwere so that they can be included in the base task but also in adapters. That way the base adapter and base task could both have the base step methods and you could just do `super().*_step` to get access to it.

flash/image/classification/model.py
```diff
@@ -87,59 +88,57 @@ def __init__(
...
             backbone_kwargs = {}
 
+        if not training_strategy_kwargs:
+            training_strategy_kwargs = {}
+
+        training_strategy_kwargs.update(
+            {
+                "ways": num_classes,
+            }
+        )
```
pietrolesci, line 107: This is a bit strange here, do we need this even when `training_strategy` is default?

flash/image/classification/adapters.py
```diff
@@ -0,0 +1,518 @@
...
+            # we are actually sampling one task per device.
+            dataset = TaskDataParallel(
+                taskset=taskset,
+                global_rank=trainer.global_rank,
+                world_size=trainer.world_size,
+                num_workers=num_workers,
+                meta_batch_size=self.meta_batch_size,
+                seed=os.getenv("PL_GLOBAL_SEED", self.seed),
+            )
+            self.trainer.accumulated_grad_batches = self.meta_batch_size / trainer.world_size
```
pietrolesci, line 241: Great!

flash/image/classification/adapters.py
```diff
@@ -0,0 +1,518 @@
...
+        Args:
+            task: Task to be used. This adapter should work with any Flash Classification task
+            backbone: Feature extractor to be used.
+            head: Predictive head.
+            algorithm_cls: Algorithm class coming
+                from: https://github.com/learnables/learn2learn/tree/master/learn2learn/algorithms/lightning
+            ways: Number of classes conserved for generating the task.
+            shots: The number of samples per label.
+            meta_batch_size: Number of task to be sampled and optimized over before doing a meta optimizer step.
+            queries: Number of support sample to be selected from the task.
```
pietrolesci, line 132: I was a bit confused here by the nomenclature of `shots` and `queries` but I confirm that it follows the l2l convention. Though, I'd like to try to summarize how I understand it in the current implementation

- `shots` is the number of samples used for adaptation
- `queries` is the number of samples used for validation
- `shots + queries` is the total number of samples per task

is this the correct interpretation?

flash/image/classification/adapters.py
```diff
@@ -0,0 +1,518 @@
...
+            queries=self.queries,
+            num_workers=num_workers,
+        )
+        if isinstance(dataset, IterableDataset):
+            shuffle = False
+            sampler = None
+        return super().process_train_dataset(
+            dataset,
+            trainer,
+            self._sanetize_batch_size(batch_size),
```
pietrolesci, line 332: EDIT: fine with this as this should be handled in the backend and at the moment l2l cannot deal with this for protonet. [Here](https://github.com/learnables/learn2learn/blob/6b741028f0812ba73da58acf7206d93654dde97a/learn2learn/algorithms/lightning/lightning_protonet.py#L115-L141) it always calls [`self.classifier.fit_`](https://github.com/learnables/learn2learn/blob/6b741028f0812ba73da58acf7206d93654dde97a/learn2learn/nn/protonet.py#L81-L100) on the support set (i.e., it is not batched). Please feel free to correct me if I'm wrong -- I'd really love having the possibility to set the batch_size within tasks so I really wish I'm wrong ahah

----

This is cleaner, however I think that having the possibility to set the `batch_size` normally in the datamodule is advantageous.

For example, assume that I specify that I want 1000 examples per class (i.e., `shots + queries = 1000`) -- I might want to do meta-learning not just in few-shots learning scenarios --, but I can process 16 examples at a time on my GPU. I would ideally want to be able to set the batch_size normally in the dataloader and automatically have the model process just 16 samples at a time in each task. In other words, ideally the process should look like

```python
for task in tasks:
    for batch in task:
        ...
```

Though, I don't know how easy it would be to implement

pietrolesci (approved):
LG%M

flash/core/classification.py
```diff
@@ -37,7 +38,29 @@ def binary_cross_entropy_with_logits(x: torch.Tensor, y: torch.Tensor) -> torch.
...
+
+        return metrics, loss_fn
+
+    def to_metrics_format(self, x: torch.Tensor) -> torch.Tensor:
+        if getattr(self.hparams, "multi_label", False):
+            return torch.sigmoid(x)
+        return torch.softmax(x, dim=1)
+
+
+class ClassificationTask(Task, ClassificationMixin):
```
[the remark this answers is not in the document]
tchaton: Yes, I believe so. Would be an awesome contribution from your side :)

flash/image/classification/model.py
tchaton: Done !

flash/image/classification/adapters.py
tchaton: Yes, updated !

.azure-pipelines/gpu-tests.yml
```diff
@@ -59,6 +59,11 @@ jobs:
         python -m coverage run --source flash -m pytest flash tests/examples/test_scripts.py -v --junitxml=$(Build.StagingDirectory)/test-results.xml --durations=30
       displayName: 'Testing'
 
+    - bash: |
+        pip install git+https://github.com/tchaton/learn2learn@flash
+        bash tests/special_tests.sh
+      displayName: 'Testing: special'
```
ethanwharris, line 65: Currently our GPU tests only run nightly. Should these special tests run every PR instead? We could temp enable them here just to at least check that everything passes

flash_examples/image_classification.py
```diff
@@ -29,7 +29,7 @@
 model = ImageClassifier(backbone="resnet18", num_classes=datamodule.num_classes)
 
 # 3. Create the trainer and finetune the model
-trainer = flash.Trainer(max_epochs=3, gpus=torch.cuda.device_count())
+trainer = flash.Trainer(max_epochs=3, gpus=torch.cuda.device_count(), limit_train_batches=2, limit_val_batches=2)
```
ethanwharris, line 32: Looks like this was an accident:
```suggestion
trainer = flash.Trainer(max_epochs=3, gpus=torch.cuda.device_count())
```

flash_examples/image_classification_meta_learning.py
```diff
@@ -0,0 +1,50 @@
...
+# 2. Build the task
+model = ImageClassifier(
+    datamodule.num_classes,
+    backbone="resnet18",
+    training_strategy="prototypicalnetworks",
+    training_strategy_kwargs={"shots": 4, "meta_batch_size": 10},
+)
+
+# 3. Create the trainer and finetune the model
+trainer = flash.Trainer(max_epochs=1, limit_train_batches=2, limit_val_batches=2)
```
ethanwharris, line 35:
```suggestion
trainer = flash.Trainer(max_epochs=1)
```

flash/image/classification/adapters.py
tchaton: The batch_size is used for the dataloader, so it should be always 1.

However, the task batch_size is defined as you shared by `shots + queries`.

To enable large values, it might be possible to perform micro batching directly within the `training_step`.

```py
embeddings = []
for chunked_task in self.chunk_task(task):
    embeddings.append(self.features(chunked_task))

...
```

But I believe there might be some better way to do this, I will think about it.

.azure-pipelines/gpu-tests.yml
tchaton: We can enable this later. I need to check with Jirka if it is worth to enable it on every PR.

flash/image/classification/adapters.py
pietrolesci: Micro-batching is very useful especially in those case in which one wants to use meta-learning to adapt a model on a downstream task without the "few-shots" limitations

flash/image/classification/adapters.py
```diff
@@ -0,0 +1,513 @@
...
+        self,
+        task: AdapterTask,
+        backbone: torch.nn.Module,
+        head: torch.nn.Module,
+        algorithm_cls: Type[LightningModule],
+        ways: int,
+        shots: int,
+        meta_batch_size: int,
+        queries: int = 1,
+        num_task: int = -1,
```
pietrolesci, line 112: @tchaton I think we might benefit from additional flexibility in the `num_task` argument. If you look at the [miniimagenet example here](https://github.com/learnables/learn2learn/blob/06893e847693a0227d5f35a6e065e6161bb08201/examples/vision/protonet_miniimagenet.py#L115-L140), you will see that they use different `num_task`'s when creating the `TaskDataset` for train, val, test

```python
train_tasks = l2l.data.TaskDataset(
    train_dataset, 
    task_transforms=train_transforms,  # note: it defaults to -1
)
test_tasks = l2l.data.TaskDataset(
    test_dataset,
    task_transforms=test_transforms,
    num_tasks=2000
)
valid_tasks = l2l.data.TaskDataset(
    valid_dataset,
    task_transforms=valid_transforms,
    num_tasks=200
)
```

flash_examples/image_classification_meta_learning.py
```diff
@@ -0,0 +1,50 @@
...
+download_data("https://pl-flash-data.s3.amazonaws.com/hymenoptera_data.zip", "./data")
+
+datamodule = ImageClassificationData.from_folders(
+    train_folder="data/hymenoptera_data/train/",
+    val_folder="data/hymenoptera_data/val/",
+)
+
+# 2. Build the task
+model = ImageClassifier(
+    datamodule.num_classes,
```
pietrolesci, line 28: Apparently, here the logic of specifying the number of classes beforehand breaks.

In particular, in the case of `training_strategy="prototypicalnetworks"` (and in general in meta-learning applications) the number of classes is only known via the `ways` in training_kwargs and can even change between training, validation, testing, and prediction. Passing a `None` currently creates a `TypeError`.

flash/image/classification/model.py
```diff
@@ -87,59 +89,58 @@ def __init__(
...
         if not backbone_kwargs:
             backbone_kwargs = {}
 
+        if not training_strategy_kwargs:
+            training_strategy_kwargs = {}
+
+        if training_strategy_kwargs != "default":
+            if "ways" in training_strategy_kwargs and training_strategy_kwargs["ways"] != num_classes:
+                raise MisconfigurationException(
+                    "When providing ways, it should match `num_classes` as mapping is not supported yet."
```
pietrolesci, line 107:
```suggestion
                    "When providing `ways` in `training_strategy_kwargs`, it should match `num_classes` as mapping is not supported yet. Specify `num_classes` to your desired value of `ways`."
```
```suggestion
                    "When providing ways, it should match `num_classes` as mapping is not supported yet."
```

pietrolesci: The proposal above is a minor revision.

A somewhat more general revision would be to always overwrite `num_classes` with `ways` and warn the user about it. What do you think?

Even more general would be to revisit the `num_classes` argument, as discussed with @ethanwharris. However, this is a larger discussion and I think it does not fit in this PR. About this, though, I'd be curious to know your (on-liner) opinion :)

flash/image/classification/adapters.py
```diff
@@ -0,0 +1,548 @@
...
+        epoch_length: Optional[int] = None,
```
pietrolesci, line 113: Missing docstring for this argument

flash/image/classification/adapters.py
```diff
@@ -0,0 +1,548 @@
...
+        test_epoch_length: Optional[int] = None,
```
pietrolesci, line 114: Missing docstring for these arguments

flash/image/classification/model.py
```diff
@@ -87,59 +89,59 @@ def __init__(
...
 
         self.save_hyperparameters()
 
         if not backbone_kwargs:
             backbone_kwargs = {}
 
+        if not training_strategy_kwargs:
+            training_strategy_kwargs = {}
+
+        if training_strategy_kwargs == "default":
```
pietrolesci, line 104: I think should be `training_strategy`

tchaton: Good catch !

flash/image/classification/adapters.py
tchaton: Done !

ethanwharris (approved):
LGTM :smiley:

requirements/datatype_image_extras.txt
```diff
@@ -2,7 +2,9 @@ matplotlib
...
```
ethanwharris, line 5:
```suggestion
```

.azure-pipelines/gpu-tests.yml
```diff
@@ -51,6 +51,7 @@ jobs:
     - bash: |
         # python -m pip install "pip==20.1"
         pip install '.[all]'
+        pip install git+https://github.com/learnables/learn2learn.git
```
ethanwharris, line 54:
```suggestion
        pip install learn2learn
```

PR #757, opened 2021-09-14
Set logo to have a white background for dark mode

SeanNaren:
Adds a white outline to the lightning flash logo which is only visible in dark mode on github.

Now it would pass a contrast checker, helpful for accessibility :)

Before:
Screenshot 2021-09-14 at 16 53 03
After:
Screenshot 2021-09-14 at 16 52 54

cc @phoebynaren who sent this to me :)


ethanwharris (approved):
LGTM :smiley:

PR #762, opened 2021-09-15
[Feat] Add ActiveLearning Loop Customization

tchaton:
## What does this PR do?

Fixes # (issue)

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/flash/image/classification/integrations/baal.py
+++ b/flash/image/classification/integrations/baal.py
@@ -1,0 +1,117 @@
+# Copyright The PyTorch Lightning team.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from copy import deepcopy
+from typing import Any, Dict, Generator, Optional
+
+import numpy as np
+import torch
+from pytorch_lightning.loops import Loop
+from pytorch_lightning.loops.fit_loop import FitLoop
+from pytorch_lightning.trainer.progress import Progress
+from pytorch_lightning.trainer.states import RunningStage, TrainerFn, TrainerStatus
+
+import flash
+from flash.core.utilities.imports import _BAAL_AVAILABLE, requires
+
+if _BAAL_AVAILABLE:
+    from baal.active import ActiveLearningDataset
+    from baal.active.heuristics import AbstractHeuristic, BALD
+
+
+class ActiveLearningTrainer(flash.Trainer):
+    def __init__(self, *args, **kwags):
+        super().__init__(*args, **kwags)
+
+        active_learning_loop = ActiveLearningLoop(label_epoch_frequency=1, heuristics=BALD())
+        active_learning_loop.connect(self.fit_loop)
+        self.fit_loop = active_learning_loop
+        active_learning_loop.trainer = self
+
+
+class ActiveLearningLoop(Loop):
+    @requires("baal")
+    def __init__(self, label_epoch_frequency: int, heuristics: "AbstractHeuristic"):
+        super().__init__()
+        self.label_epoch_frequency = label_epoch_frequency
+        self.fit_loop: Optional[FitLoop] = None
+        self.state_dict: Optional[Dict[str, torch.Tensor]] = None
+        self.progress = Progress()
+        self.heuristics = heuristics
+        self._should_continue: bool = False
+
+    def __getattr__(self, key):
+        if key not in self.__dict__:
+            return getattr(self.fit_loop, key)
+        return self.__dict__[key]
+
+    def connect(self, fit_loop: FitLoop):
+        self.fit_loop = fit_loop
+        self.max_epochs = self.fit_loop.max_epochs
+        self.fit_loop.max_epochs = self.label_epoch_frequency
+
+    def _to_active_dataset(self, dataset):
+        return ActiveLearningDataset(dataset, labelled=torch.arange(len(dataset) - 10))
+
+    def on_run_start(self, *args: Any, **kwargs: Any) -> None:
+        self.trainer.predict_loop._return_predictions = True
+        self.state_dict = deepcopy(self.trainer.lightning_module.state_dict())
+        if self.trainer.datamodule is not None:
+            self.trainer.datamodule.register_dataset_hooks(self._to_active_dataset, RunningStage.TRAINING)
+            self.trainer.datamodule.register_dataset_hooks(self._to_active_dataset, RunningStage.VALIDATING)
+            self.trainer.datamodule.register_dataset_hooks(self._to_active_dataset, RunningStage.TESTING)
+            self.trainer.datamodule.register_dataset_hooks(self._to_active_dataset, RunningStage.PREDICTING)
+
+    def on_advance_start(self, *args: Any, **kwargs: Any) -> None:
+        self.trainer.train_dataloader = None
+        self.trainer.val_dataloaders = None
+        self.trainer.lightning_module.train_dataloader.unpatch(self.trainer.lightning_module)
+        self.trainer.lightning_module.train_dataloader.unpatch(self.trainer.lightning_module)
+        self.trainer.reset_train_val_dataloaders(self.trainer.lightning_module)
+        self.trainer.reset_predict_dataloader(self.trainer.lightning_module)
+
+    def advance(self, *args: Any, **kwargs: Any) -> None:
+        self.progress.increment_started()
+        self.fit_loop.run()
+        self._reset_predicting()
+        predictions = self.trainer.predict_loop.run()
+        self.heuristics(self._generator_prediction(predictions[0]))
+        breakpoint()
+        self._reset_training()
+        self.progress.increment_processed()
+
+    def on_advance_end(self) -> None:
+        self.trainer.lightning_module.load_state_dict(self.state_dict)
+        self.progress.increment_completed()
+        return super().on_advance_end()
+
+    @property
+    def done(self) -> bool:
+        return self.progress.current.completed > self.max_epochs or self._should_continue
+
+    def reset(self) -> None:
+        pass
+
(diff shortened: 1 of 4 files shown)


flash/image/classification/integrations/baal.py
```diff
@@ -0,0 +1,143 @@
...
+            return getattr(self.fit_loop, key)
+        return self.__dict__[key]
+
+    def connect(self, fit_loop: FitLoop):
+        self.fit_loop = fit_loop
+        self.max_epochs = self.fit_loop.max_epochs
+        self.fit_loop.max_epochs = self.label_epoch_frequency
+
+    def _to_active_dataset(self, dataset):
+        return ActiveLearningDataset(dataset, labelled=torch.arange(len(dataset) - 10))
```
parmidaatg, line 67: is there a reason that we force what percentage of the data is labelled?

flash/image/classification/integrations/baal.py
```diff
@@ -0,0 +1,143 @@
...
+
+    def on_run_start(self, *args: Any, **kwargs: Any) -> None:
+        self.trainer.predict_loop._return_predictions = True
+        self.state_dict = deepcopy(self.trainer.lightning_module.state_dict())
+        if self.trainer.datamodule is not None:
+            self.trainer.datamodule.register_dataset_hooks(self._to_active_dataset, RunningStage.TRAINING)
+            self.trainer.datamodule.register_dataset_hooks(self._to_active_dataset, RunningStage.VALIDATING)
+            self.trainer.datamodule.register_dataset_hooks(self._to_active_dataset, RunningStage.TESTING)
+            self.trainer.datamodule.register_dataset_hooks(self._to_active_dataset, RunningStage.PREDICTING)
```
parmidaatg, line 76: normally, we do not wrap the data used for validation and test with `ActiveLearningDataset`, as these sets are supposed to be fully labelled. Is there a reason that we wrap all of them with `ActiveLearningDataset` here? 
On the other hand, Do we expect the prediction set to be the `pool` or separate?

parmidaatg (commented):
Thank you so much for this work. It is very well done and clean. 
I need to do another pass just to make sure I am not missing anything but I put two comments to understand the mechanics better. 
the main question that I have is that where would we use `patch_module`  to make sure that the model is changed to do MCD (https://github.com/ElementAI/baal/blob/master/src/baal/active/__init__.py#L9)?

flash/image/classification/integrations/baal.py
pietrolesci: To add to @parmidaatg's comment, it would be great to have the option to turn-on "dynamic" validation set: the validation set contains the same number of instances as the current training budget. This mimics a more realistic active learning scenario

flash/image/classification/integrations/baal.py
tchaton: This is a hack while the PR is in progress. 
The length is being checked and it would break if no labelled data are provided. 
This would definitely be changed to the right approach following your example / codebase.

I was actually thinking to do the following:

```py
while not done:
    reset model
    train on train_dataset
    compute scores on predict train_dataset
    request for labelisation
    merge train_dataset + predict labelled data
    predict_dataset : Subset of non label data
    store current labellisation state in the loop.

move labelled data to train folder
```

tchaton: If you believe another approach should be taken, please share the pseudo code and I will implement it :) 

Enabling non labelled dataset for training might requires more work on the data loading part, but definitely feasible. 

Option 1:

Provide a way for the datamodule to load unlabelled data directly

```py
ImageClassificationData.from_folders(
    train_folder=...
    trainer_non_labelled_folder=...
)
```

Option 2:

Add support for unlabelled folder with contains non-labelled images.

I would be curious to hear the best practices on data management for active learning.

flash/image/classification/integrations/baal.py
tchaton: Thanks @parmidaatg @pietrolesci,

@parmidaatg Great point, there is no reason for them to get wrapped. I was just debugging the code :)
Only predict_dataset would be wrapped into an ActiveLearningDataset as it contains the non-labelled data.

@pietrolesci I would be curious if you could share more details on how dynamic validation set is usually implemented.

tchaton:
Dear @parmidaatg,

> the main question that I have is that where would we use patch_module to make sure that the model is changed to do MCD

This would be applied right before running the predict loop. 
Right now, I am patching the `predict_step` to add mc evaluation 
and manually setting dropout to train mode.

But relying on the Baal internals is definitely my goal.

One issue I faced was: LightningModule can't be wrapped with MCDropoutModule / ModelWrapper.
I will look into the `patch_module`.

Best,
T.C

flash/image/classification/integrations/baal.py
pietrolesci: @tchaton I think that on the datamodule side things should remain the same, from an API perspective. Internally, the class should split the training set into a pool and a labelled set. That's usually what one expects.

Maybe the process can look like
```
while not done
    if labelled data:  # <- in the first iteration, if you have data train otherwise use the initialized model
        train on labelled
    predict the pool dataset
    compute scores
    etc   
```
Usually (even if tiny and chosen at random) a labelled budget is the starting point of active learning. However, having the possibility of allowing the model to choose the instances to label from the very beginning can be interesting (think of pretrained models that you want to adapt to downstream task: they might know what they want)

flash/image/classification/integrations/baal.py
pietrolesci: @tchaton one way I did it in the past was to add to the validation set the same number of instances added to the training set, chosen at random.

I must add, in the future one can also think to devise better strategies to actively select even the validation set (think of the applications in meta-learning), but this is another discussion :)

pietrolesci: @tchaton to address your reply to @parmidaatg - do you think there is a clean way to allow the user to choose what to wrap?

tchaton:
Hey @parmidaatg,

After chatting with @pietrolesci, we ended up with this API.

```py
# production use case - from scratch

ImageClassificationData.from_folders(
    train_folders="path_to_unlabelled_data", # none, needs to come from the model
)

trainer = ActivateLearningTrainer(
    val_size = 0.1,
)

# production use case - with little data

ImageClassificationData.from_folders(
    train_folders=["path_to_labelled_data", "path_to_unlabelled_data"], # num_classes directly, none
)

trainer = ActivateLearningTrainer(
    val_size = 0.1,
)

# research use case - entirely labelled dataset

ImageClassificationData.from_folders(
    train_folders=path_to_labelled_data, # num_classes directly
)

def on_run_start(dataset) -> Tuple[Optional[Dataset], Optional[Dataset]]:
    return None, predict_data


def labelisation_stopping_criteria(labels, num_classes):
    return np.bicount(labels) > 5

trainer = ActivateLearningTrainer(
    val_size = 0.1, callable
    on_run_start = on_run_start,
    heuristic: ... compute the scores
    labelisation_stopping_criteria = labelisation_stopping_criteria # used for meta-learning with active learning
)
```

What are your thoughts ?

flash/image/classification/integrations/baal.py
parmidaatg: I agree with @pietrolesci  on this. I think, prior to training it is good to give the option to the user to input which samples are labelled. Otherwise, they would use the `label_randomly` functionality of `ActiveDataset` to label a few samples and then start the training. It totally makes sense for the training to not support an empty training set (no labels in the dataset). `ActiveDataset` is automatically taking care of what is used for the training and what is used for the `pool` so that part doesn't need handling unless you would need to update another dataset with the newly created `ActiveDataset.pool` at each step.

flash/image/classification/integrations/baal.py
parmidaatg: true, we haven't supported the automated selection of validation-set in BaaL yet. I believe it should be an open issue which is not a priority at the time but potentially it would be nice to support having the same number of examples in validation set or maybe the same potion as the labelled data vs pool.

parmidaatg:
> Hey @parmidaatg,
> 
> After chatting with @pietrolesci, we ended up with this API.
> 
> ```python
> # production use case - from scratch
> 
> ImageClassificationData.from_folders(
>     train_folders="path_to_unlabelled_data", # none, needs to come from the model
> )
> 
> trainer = ActivateLearningTrainer(
>     val_size = 0.1,
> )
> 
> # production use case - with little data
> 
> ImageClassificationData.from_folders(
>     train_folders=["path_to_labelled_data", "path_to_unlabelled_data"], # num_classes directly, none
> )
> 
> trainer = ActivateLearningTrainer(
>     val_size = 0.1,
> )
> 
> # research use case - entirely labelled dataset
> 
> ImageClassificationData.from_folders(
>     train_folders=path_to_labelled_data, # num_classes directly
> )
> 
> def on_run_start(dataset) -> Tuple[Optional[Dataset], Optional[Dataset]]:
>     return None, predict_data
> 
> 
> def labelisation_stopping_criteria(labels, num_classes):
>     return np.bicount(labels) > 5
> 
> trainer = ActivateLearningTrainer(
>     val_size = 0.1, callable
>     on_run_start = on_run_start,
>     heuristic: ... compute the scores
>     labelisation_stopping_criteria = labelisation_stopping_criteria # used for meta-learning with active learning
> )
> ```
> 
> What are your thoughts ?

This is great. Some thoughts:
1-  in the research use case, is there a parameter accessible via `ActivateLearningTrainer` by which the user defines how much of the data should be used as `pool`? 
2- for the production use case, is it tangible to implement a `label` function so that the user could add their input label to the dataset? `ActiveDataset` only updates the labelled portion, if this functionality is available. Maybe take a look at Baal's `FileDataset` wrapper -> https://github.com/ElementAI/baal/blob/master/src/baal/active/file_dataset.py
3-  Is the first use case too I think if the user wants to benefit from ActiveLearning they would need to be able to add the newly `Labelled` data in training and finetune at each step based on that, what do you think?

I am really excited for this integration :) thanks a lot for the time you put in this

tchaton:
Dear @parmidaatg @pietrolesci,

As the ActiveLoop is started to look slightly more hacky, we are thinking to introduce an ActiveLearningCombinor. I will investigate the following.

```py
labelled_dm = ImageClassificationData.from_folders(train_folder="path_to_unlabelled_data")
unlabelled_dm = ImageClassificationData.from_folders(train_folder="path_to_labelled_data")
dm = ActiveLearningDataModuleCombinor(labelled=labelled_dm, unlabelled=unlabelled_dm)
```

Best,
T.C

tchaton:
Moved the PR to https://github.com/PyTorchLightning/lightning-flash/pull/779

PR #777, opened 2021-09-20
[WIP] PoC: Revamp optimizer and scheduler experience using registries

### LINKED ISSUE #752: [RFC] Enable Lightning Flash to support scheduler on step or monitor
## 🐛 Bug

Currently, in Lightning Flash, 

```py
class Task(...):

    def configure_optimizers(self) -> Union[Optimizer, Tuple[List[Optimizer], List[_LRScheduler]]]:
        optimizer = self.optimizer
        if not isinstance(self.optimizer, Optimizer):
            self.optimizer_kwargs["lr"] = self.learning_rate
            optimizer = optimizer(filter(lambda p: p.requires_grad, self.parameters()), **self.optimizer_kwargs)
        if self.scheduler:
            return [optimizer], [self._instantiate_scheduler(optimizer)]
        return optimizer
```

However, in PyTorch Lightning, someone would have to do the follow to enable training on step or monitoring for a scheduler

```py
class Task(...):


    def configure_optimizers(self) -> Union[Optimizer, Tuple[List[Optimizer], List[_LRScheduler]]]:
        optimizer = self.optimizer
        if not isinstance(self.optimizer, Optimizer):
            self.optimizer_kwargs["lr"] = self.learning_rate
            optimizer = optimizer(filter(lambda p: p.requires_grad, self.parameters()), **self.optimizer_kwargs)
        if self.scheduler:
            return [optimizer], [{"scheduler": self._instantiate_scheduler(optimizer), "interval": "step", "monitor": "val_loss"}]
        return optimizer
```

Here is the default scheduler dict.

```py
    return {
        "scheduler": None,
        "name": None,  # no custom name
        "interval": "epoch",  # after epoch is over
        "frequency": 1,  # every epoch/batch
        "reduce_on_plateau": False,  # most often not ReduceLROnPlateau scheduler
        "monitor": None,  # value to monitor for ReduceLROnPlateau
        "strict": True,  # enforce that the monitor exists for ReduceLROnPlateau
        "opt_idx": None,  # necessary to store opt_idx when optimizer frequencies are specified
    }
```

A possible solution API to support scheduler on step would be to provide the entire default dict while instantiating as follow.

```py
ImageClasifier(
    scheduler={
        "scheduler": torch.optim.lr_scheduler.StepLR,
        "interval": "step",
    },
    scheduler_kwargs={"step_size": 1},
)

or 

```py
ImageClasifier(
    scheduler={
        "scheduler": torch.optim.lr_scheduler.ReduceLROnPlateau
        "monitor": "val_loss",
    },
    scheduler_kwargs={"step_size": 1},
)
```

### To Reproduce

2. Run '....'

### Environment

 - PyTorch Version (e.g., 1.0):
 - OS (e.g., Linux):
 - How you installed PyTorch (`conda`, `pip`, source):
 - Build command you used (if compiling from source):
 - Python version:
 - CUDA/cuDNN version:
 - GPU models and configuration:
 - Any other relevant information:

karthikrangasai:
## What does this PR do?

Fixes #752 

Optimizer and scheduler can currently be an instance but that doesn't comply with the way the code works since you need the model to the instantiate optimizer and need the optimizer to instantiate the scheduler. The main idea is to use the `FlashRegistry` class for this.

This PR focuses on fixing this issue using a few ideas:
- [ ] To pass a partial function which instantiates the optimizer or scheduler given the object it needs to wrap
- [ ] Use the scheduler registry: So in theory you can select a scheduler with a string.

Some ideas on how we can use the registry is mentioned in the issue #752 

### Configurable scheduler registry (help needed):

Most schedulers have some non-default parameters that need to input during instantiation. The idea here is to be able to configure the scheduler instantiation just using the string. For example with a `StepLR` we should be able to do:
```python
model = XYZ_Task(backbone, optimizer="Adam", scheduler="StepLR--10")
```

For this, I am planning to inspect the scheduler's `__init__` method for the parameters and it's types. The string will have the arguments in the same order and we can split on the `--` and typecast all the arguments.

### On step or Monitor for LR

Still in progress ...

## Before submitting

## PR review

## Did you have fun?


tchaton:
Dear @karthikrangasai,

I added some modifications. Still wip.

Do you think we should remove entirely `optimizer_kwargs` and `scheduler_kwargs` ? 
Adding support for a tuple seems slightly counter-intuive as users might want to control arguments through keyword arguments.

Best,
T.C

tchaton:
Hey @SeanNaren @ethanwharris.

Should we drop `learning_rate` from the Tasks and force kwargs directly ?

Best,
T.C

PR #779, opened 2021-09-20
[Feat] Add ActiveLearning Loop Customization v2

### LINKED ISSUE #761: [RFC] Feature: support Active Learning training loop
# Feature 🚀: Active Learning training loop

Active learning is a special case of machine learning in which a learning algorithm is able to
interactively query the user (or some other information source) to obtain the desired outputs
at new data points.

In practice, it is often implemented as "pool-based" active learning: data is readily available
but labelling is costly. The active learning process proceeds as follows. Assuming you have a 
unlabelled dataset (a "pool") and you labelled (or "acquired") (maybe randomly) some inputs from the pool

1. Train the model for one epoch on those labelled data (within each epoch the data can be seen multiple
times - this is important especially when the initial labelled dataset is small; of course how
many times the dataset is repeated within each epoch can vary through the active training)

1. Evaluate the model on the pool -> (computationally intensive, maybe use the new `torch.inference_mode()`
context manager? Maybe use something like [toma](https://github.com/BlackHC/toma) to max out memory space?)

1. Feed the model predictions/statistics (depends on the specific active learning strategy) to
an "acquisition"/__scoring function__ which scores the inputs in the pool

1. Once the scores are available, apply a __query strategy__ (e.g., top _k_, max margin, least confidence)
that selects the actual inputs to label

1. Label inputs from the pool and add it to the training set -> (might be problematic in distributed setting)

1. (Optional) Re-initialize the model from scratch (weights, scheduler, optimizer)

In the current state of Flash, this can be achieved via a callback. However, this poses challenges
in the distributed setting, makes it difficult to reset states, and makes it more difficult to work
with existing `training_strategies`.

## Proposed API
The alternative is to implement active learning in Flash via a new trainer, the `ActiveLearningTrainer`.
This new trainer would provide access to the `ActiveLearningLoop` ... [here is for you @tchaton :)]

This is justified by the fact that

- Active learning is effectively an outer loop on top of the standard
training loop

- User don't have to learn a new API as it will be the same as the current trainer

- Researchers can easily hack into the active learning loop

## Desiderata

From a researcher point of view, it is important to be able to hack the following things

- How many times the training data are repeated within each epoch and be able to change that, much like a scheduler

- Change how the model is evaluated on the pool (e.g., certain AL strategies need many forward passes on the same
data to estimate uncertainty)

- Change the acquisition (aka scoring) functions

- Change the query strategy

tchaton:
## What does this PR do?

This PR integrates Baal within Flash to offer ActiveLearning Training.

Fixes #761

## Before submitting

## PR review

## Did you have fun?


flash/image/classification/model.py
```diff
@@ -118,6 +118,7 @@ def __init__(
 
         head = head(num_features, num_classes) if isinstance(head, FunctionType) else head
         head = head or nn.Sequential(
+            nn.Dropout(p=0.1),
```
ethanwharris, line 121: I don't think we should add this here. The classification head just replaces the last layer, so if there was already dropout before it then you would now get: dropout -> dropout -> linear.

flash/image/classification/integrations/baal.py
```diff
@@ -0,0 +1,250 @@
...
+        heuristic: "AbstractHeuristic" = BALD(reduction=np.mean),
+        map_dataset_to_labelled: Optional[Callable] = dataset_to_non_labelled_tensor,
+    ):
+        self.labelled = labelled
+        self.unlabelled = unlabelled
+        self.heuristic = heuristic
+        self.map_dataset_to_labelled = map_dataset_to_labelled
+
+        if self.unlabelled:
+            raise MisconfigurationException("The unlabelled `datamodule` isn't support yet.")
```
ethanwharris, line 60: If the argument is not supported it might be cleaner to just remove it

ethanwharris (commented):
Looks good, could we have some documentation?

flash/image/classification/model.py
tchaton: This is quite un-likely no ? The dropout are usually place within head or deep within the model.

ethanwharris: Well, most models don't use dropout at all. So you would now be getting a different network architecture. In any case this means that results obtained with the actually architecture cannot be reproduced within Flash.

flash/image/classification/integrations/baal.py
tchaton: This is not supported, but I would like to work on this asap, so we support the production use case. The problem is that I need to label data ahah :)
I plan to integrate with LabelStudio, so I can do this from the UI.

tchaton: Ok, removed it and I will add it back when supported.

flash/image/classification/model.py
tchaton: Removed !

flash/image/classification/integrations/baal/__init__.py
```diff
@@ -0,0 +1,31 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import flash
+from flash.core.utilities.imports import requires
+from flash.image.classification.integrations.baal.loop import ActiveLearningDataModule, ActiveLearningLoop  # noqa F401
+
+
+class ActiveLearningTrainer(flash.Trainer):
+    @requires("baal")
+    def __init__(self, *args, label_epoch_frequency: int = 1, inference_iteration: int = 2, **kwags):
```
awaelchli, line 21: should probably not support positional args :)

flash/image/classification/integrations/baal/__init__.py
```diff
@@ -0,0 +1,31 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import flash
+from flash.core.utilities.imports import requires
+from flash.image.classification.integrations.baal.loop import ActiveLearningDataModule, ActiveLearningLoop  # noqa F401
+
+
+class ActiveLearningTrainer(flash.Trainer):
```
awaelchli, line 19: is this just here to make the example shorter or are you saying that loop customization should be packaged like this in general?

flash/image/classification/integrations/baal/loop.py
```diff
@@ -0,0 +1,141 @@
...
+            _PatchDataLoader(getattr(self.trainer.datamodule, dataloader_name)(), running_state),
+        )
+        setattr(self.trainer, dataloader_name, None)
+        getattr(self.trainer, f"reset_{dataloader_name}")(self.trainer.lightning_module)
+
+    def _enable_mc_dropout(self):
+        # prevent the model to put into val model - hack
+        self.trainer.lightning_module.on_predict_model_eval = self._do_nothing
+        for _, module in self.trainer.lightning_module.named_modules():
+            if isinstance(module, _BatchNorm):
```
awaelchli, line 143: this seems very arbitrary. I would be interested why batch norm in particular and not other types. do you have any resources on this?

awaelchli (approved):

flash/image/classification/integrations/baal/loop.py
```diff
@@ -0,0 +1,162 @@
...
+                module.eval()
+        # save the predict_step and replace it within a mc_inference function
+        if getattr(self.trainer.lightning_module, "_predict_step", None) is None:
+            self.trainer.lightning_module._predict_step = self.trainer.lightning_module.predict_step
+        self.trainer.lightning_module.predict_step = partial(
```
awaelchli, line 148: isn't this its own loop? 

In Lightning, replacing a users methods would be unexpected behavior. Is this done often in flash?

flash/image/classification/integrations/baal/__init__.py
tchaton: That's a great and open question. I think it is simpler to package it this way in Flash to make example shorter.

flash/image/classification/integrations/baal/loop.py
tchaton: Not it is specific to the LightningModule, but it is currently the only option to enable running the predict_step multuple times and aggregating the predictions to perform MCDropout evaluation.

flash/image/classification/integrations/baal/__init__.py
tchaton: Yes, I don't think the Trainer as any anyway :)

flash/image/classification/integrations/baal/__init__.py
justusschock: I strongly think we should not do it that way! Whatever we do here is likely to be used by others as well. I think we need to introduce a formal API for exchanging loops. IMO if we need to subclass the trainer (which is supposed to be the fundamental element of PL) for something we also want to support in PL in general, it shows that we might need a better API there (or in this case: An API at all).

tchaton: Fine with me ! Let me make it explicit within the example how we connect loops.

flash/image/classification/integrations/baal/__init__.py
tchaton: Removed !

flash/image/classification/integrations/baal/data.py
```diff
@@ -0,0 +1,121 @@
...
+
+    def train_dataloader(self) -> "DataLoader":
+        self.labelled._train_ds = self._dataset
+        return self.labelled.train_dataloader()
+
+    def predict_dataloader(self) -> "DataLoader":
+        self.labelled._train_ds = self._dataset.pool
+        return self.labelled.train_dataloader()
+
+    def label(self, predictions: Any = None, indices=None):
```
Dref360, line 106: We do have `ActiveLearningDataset.label_randomly` which is useful in research to kickstart the process. Might be useful here too.
https://baal.readthedocs.io/en/latest/api/index.html#baal.active.ActiveLearningDataset.label_randomly

flash/image/classification/integrations/baal/data.py
```diff
@@ -0,0 +1,121 @@
...
+        if predictions and indices:
+            raise MisconfigurationException(
+                "The `predictions` and `indices` are mutually exclusive, pass only of one them."
+            )
+        if predictions:
+            uncertainties = [self.heuristic.get_uncertainties(np.asarray(p)) for idx, p in enumerate(predictions)]
```
Dref360, line 112: Can we call `self.heuristic.get_uncertainties(predictions)` directly?

flash/image/classification/integrations/baal/loop.py
```diff
@@ -0,0 +1,162 @@
...
+                    uncertainties = heuristic(predictions)
+                    request labellelisation for the sample with highest uncertainties under a given budget
+
+        Args:
+            label_epoch_frequency: Number of epoch to train on before requesting labellisation.
+            inference_iteration: Number of inference to perform to compute uncertainty.
+        """
+        super().__init__()
+        self.label_epoch_frequency = label_epoch_frequency
+        self.inference_iteration = inference_iteration
```
Dref360, line 54: This would be used during testing too? ie. would we be using MC-Dropout to get test loss as well?

flash/image/classification/integrations/baal/loop.py
```diff
@@ -0,0 +1,162 @@
...
+        self.fit_loop: Optional[FitLoop] = None
+        self.progress = Progress()
+        self._should_continue: bool = False
+        self._model_state_dict: Optional[Dict[str, torch.Tensor]] = None
+
+    @property
+    def done(self) -> bool:
+        return self.progress.current.completed > self.max_epochs or self._should_continue
```
Dref360, line 62:
```suggestion
        return self.progress.current.completed > self.max_epochs or not self._should_continue
```

flash/image/classification/integrations/baal/loop.py
```diff
@@ -0,0 +1,162 @@
...
+            inference_iteration=self.inference_iteration,
+        )
+
+    def _do_nothing(self, *args, **kwargs):
+        pass
+
+    @staticmethod
+    def _mc_inference(*args, predict_step_fn: Callable = None, inference_iteration: int = None, **kwargs):
+        out = []
+        for _ in range(inference_iteration):
```
Dref360, line 160: BaaL expects a shape [num_samples, num_classes, num_iterations]

Dref360 (changes requested):
Some comments, awesome work!

flash/image/classification/integrations/baal/data.py
tchaton: Yes, I am going to move it directly within `InferenceMCDropoutTask`, so we reduce memory usage.

flash/image/classification/integrations/baal/loop.py
tchaton: Done ! Good catch !

flash/image/classification/integrations/baal/loop.py
tchaton: Good catch !

tchaton: I tried to use it this way, it seems to break

flash/image/classification/integrations/baal/loop.py
tchaton: I removed it as it is not used right now.

flash/image/classification/integrations/baal/data.py
tchaton: Added the option.

flash/image/classification/integrations/baal/loop.py
tchaton: Add supported for validation. Without the MC-Dropout as it is not used to compute uncertainties but just model metrics.

flash/image/classification/integrations/baal/data.py
```diff
@@ -0,0 +1,161 @@
...
+            raise MisconfigurationException("The labelled `datamodule` should be provided.")
+
+        if not self.labelled.num_classes:
+            raise MisconfigurationException("The labelled dataset should be labelled")
+
+        if self.labelled and (
+            self.labelled._val_ds is not None
+            or self.labelled._test_ds is not None
+            or self.labelled._predict_ds is not None
+        ):
```
pietrolesci, line 95: I agree with the API here, i.e. force the user to provide only the train dataset. 

However, for the "research case" it would be nice to be able to provide the test dataset as well. This would allow us to easily test within the active learning loop and create the common "training budget vs test accuracy" graphs that are often reported in papers (example attached below).

What do you think?


active-learning-fig2
)

flash/image/classification/integrations/baal/dropout.py
```diff
@@ -0,0 +1,43 @@
...
+
+    def predict_step(self, batch, batch_idx, dataloader_idx: int = 0):
+        out = []
+        for _ in range(self.inference_iteration):
+            out.append(self.parent_module.predict_step(batch, batch_idx, dataloader_idx=dataloader_idx))
+        # BaaL expects a shape [num_samples, num_classes, num_iterations]
+        uncertainties = [
+            self.heuristic.get_uncertainties(torch.tensor(o).unsqueeze(0)) for o in torch.tensor(out).permute((1, 2, 0))
+        ]
+        return uncertainties
```
pietrolesci, line 43: Nice!

Just a note: In the future, we might want to be able to work here and make prediction as fast as possible (e.g., biggest batch size possible, onnx?)

pietrolesci (approved):
I left some comments regarding the test dataset in the `ActiveLearningDataModule`

docs/source/integrations/baal.rst
```diff
@@ -0,0 +1,33 @@
...
+
+.. raw:: html
+
+  <div style="margin-top: 20px; margin-bottom: 20px">
+    <img src="https://raw.githubusercontent.com/ElementAI/baal/master/docs/_static/images/logo-transparent.png" width="100px">
+  </div>
+
+
+Active Learning is an AI sub-field focusing on adding human in the loop.
+The most uncertain samples will be labelled to accelerate model training cycle.
```
SeanNaren, line 18:
```suggestion
Active Learning is a sub-field in AI, focusing on adding a human in the learning loop.
The most uncertain samples will be labelled by the human to accelerate model training cycle.
```

flash/image/classification/integrations/baal/data.py
```diff
@@ -0,0 +1,162 @@
...
+
+def dataset_to_non_labelled_tensor(dataset: BaseAutoDataset) -> torch.tensor:
+    return torch.zeros(len(dataset))
+
+
+def filter_unlabelled_data(dataset: BaseAutoDataset) -> Dataset:
+    return dataset
+
+
+def train_val_split(dataset: Dataset, val_size=0.1):
```
SeanNaren, line 48:
```suggestion
def train_val_split(dataset: Dataset, val_size:float=0.1):
```

flash/image/classification/integrations/baal/data.py
SeanNaren: Fully agree with @pietrolesci here to prove out the active learning case! maybe this can be a separate PR?

SeanNaren (approved):
Crushing it @tchaton!

docs/source/integrations/baal.rst
tchaton: Done !

flash/image/classification/integrations/baal/data.py
tchaton: Done !

docs/source/integrations/baal.rst
```diff
@@ -0,0 +1,33 @@
...
+
+.. raw:: html
+
+  <div style="margin-top: 20px; margin-bottom: 20px">
+    <img src="https://raw.githubusercontent.com/ElementAI/baal/master/docs/literature/images/Baalscheme.svg" width="400px">
+    <p align="center">Credit to ElementAI / Baal Team for creating this diagram flow</p>
+    <br />
+  </div>
+
+With its integration within Flash, Active Learning process is made even simpler than before.
```
ethanwharris, line 28:
```suggestion
With its integration within Flash, the Active Learning process is simpler than ever before.
```

flash/image/classification/integrations/baal/data.py
```diff
@@ -0,0 +1,162 @@
...
+    ):
+        """The `ActiveLearningDataModule` handles data manipulation for ActiveLearning.
+
+        Args:
+            labelled: DataModule containing labelled train data for research use-case.
+                The labelled data would be masked.
+            heuristic: Sorting algorithm used to rank samples on how likely they can help with model performance.
+            map_dataset_to_labelled: Function used to emulate masking on labelled dataset.
+            filter_unlabelled_data: Function used to filter the unlabelled data while computing uncertainties.
+            num_label_randomly: Number of samples to randomly labelled from the uncertaintie scores
```
ethanwharris, line 74:
```suggestion
            num_label_randomly: Number of samples to randomly label from the uncertainty scores.
```

flash/image/classification/integrations/baal/data.py
tchaton: Done !

ethanwharris (approved):
LGTM

pietrolesci (approved):
LGTM

Great work @tchaton!

flash/image/classification/integrations/baal/dropout.py
tchaton: Done, added the computation on epoch end. It should be faster :)

docs/source/integrations/baal.rst
tchaton: Done !

flash/image/classification/integrations/baal/data.py
tchaton: Done !

tchaton:
> LGTM
> 
> Great work @tchaton!

Thanks @pietrolesci for helping understand what was required :) Great team work !

Next step, let's reproduce the result from Baal, LabelStudio integration.

Best,
T.C

PR #815, opened 2021-09-29
[Refactor] Improve preprocess / data_source relation

### LINKED ISSUE #778: Preprocessing module is very confusing
## ❓ Questions and Help

There seems to be a bit of code smell cropping up with the #770  issue related to `to_onnx` and how example inputs are used. I have a few questions.

#### What is your question?

1. How does the DataPipeline class relate to the preprocess module? The documentation for this class is very sparse.
2. Why is it the responsibility of the Preprocessor to instantiate the data source when `from_data_source("something custom")` is called to create a data module? This flow is very unintuitive.
3. `ApplyToKeys` operates on dictionaries, and `example_input_array` can be a tensor, tuple, or dict. Where and how is this handled? For example in `to_onnx` or `summarize` the example_input_array tensor is passed through the model, which works fine on just the model, but the preprocessing expects a `dict` for the most part since it is highly encouraged in the docs to use that function as everything in flash is supposed to be a dict until right before inference. This results in all kinds of strange errors. 

#### Code

The fix for 3 seems to be to do this on my custom `Task`
```
    def _apply_batch_transfer_handler(self, batch: Any, device: Optional[torch.device] = None, dataloader_idx: Optional[int] = None) -> Any:
        if isinstance(batch, torch.Tensor):
            return super()._apply_batch_transfer_handler(batch={DefaultDataKeys.INPUT: batch}, device=device, dataloader_idx=dataloader_idx)[DefaultDataKeys.INPUT]
        else:
            return super()._apply_batch_transfer_handler(batch, device=device, dataloader_idx=dataloader_idx)
```

But I have no idea what kinds of side effects this would have and it feels like a strange fix to a fundamental problem.

#### What have you tried?

This issue sprang up when I implemented a custom preprocess class and spiraled from there. `to_onnx` and `summarize` (by extension, training) are broken due to this bug. The ImageClassificationPreprocess won't work for me because my outputs are continuous,

#### What's your environment?

 - OS: All
 - Packaging: conda/pip
 - Version lightning 1.4.7, flash 0.5.0, bolts 0.4.0

tchaton:
## What does this PR do?

Currently, the Processing part of Flash is quite confusing.
This is due the preprocess owning the data sources and de-serializer.

This PR introduces a DataSourceCollection which would be its own object and be used to create dataset for the users who might want to opt-out of Flash.

But this also makes DataModule code more readable as more sequential.

API

```py
    loader = ImageClassificationDataSourceCollection()

    train_ds, val_ds, test_ds, predict_ds = loader.from_csv(
        "image",
        "target",
        train_file=single_target_csv,
    )

    assert train_ds
    assert not val_ds
    assert not test_ds
    assert not predict_ds

    dm = ImageClassificationDataV2.from_csv(
        "image",
        "target",
        train_file=single_target_csv,
        predict_file=single_target_csv,
        batch_size=2,
        num_workers=0,
        train_transform=default_transforms((256, 256)),
        predict_transform=default_transforms((256, 256)),
    )

    batch = next(iter(dm.train_dataloader()))
    assert batch
```

Example for the ImageClassificationDataSourceCollection

```py
class ImageClassificationDataSourceCollection(DataSourceCollection):
    def __init__(self, **data_source_kwargs):
        super().__init__(
            data_sources={
                DefaultDataSources.FIFTYONE: ImageFiftyOneDataSource(**data_source_kwargs),
                DefaultDataSources.FILES: ImagePathsDataSource(),
                DefaultDataSources.FOLDERS: ImagePathsDataSource(),
                DefaultDataSources.NUMPY: ImageNumpyDataSource(),
                DefaultDataSources.TENSORS: ImageTensorDataSource(),
                DefaultDataSources.DATAFRAME: ImageClassificationDataFrameDataSource(),
                DefaultDataSources.CSV: ImageClassificationDataFrameDataSource(),
                DefaultDataSources.LABELSTUDIO: LabelStudioImageClassificationDataSource(),
            },
            default_data_source=DefaultDataSources.FILES,
        )


class ImageClassificationDataV2(DataModule):

    preprocess_cls = ImageClassificationPreprocessV2
    data_source_collection_cls = ImageClassificationDataSourceCollection
    deserializer_cls = ImageDeserializer
```

Here is the class_method from the DataModule. As one can see, the data_source_collection is used to extract the associated data_soruce and create his datasets.

```py
        data_source_collection = cls.data_source_collection_cls(**preprocess_kwargs)
        data_source: ImageClassificationDataFrameDataSource = data_source_collection[DefaultDataSources.CSV]
        train_ds, val_test, test_ds, predict_ds = data_source.from_csv(
            input_field=input_field,
            target_fields=target_fields,
            train_file=train_file,
            train_images_root=train_images_root,
            train_resolver=train_resolver,
            val_file=val_file,
            val_images_root=val_images_root,
            val_resolver=val_resolver,
            test_file=test_file,
            test_images_root=test_images_root,
            test_resolver=test_resolver,
            predict_file=predict_file,
            predict_images_root=predict_images_root,
            predict_resolver=predict_resolver,
        )

        return cls.from_data_source(
            data_source,
            data_source_collection,
            train_ds,
            val_test,
            test_ds,
            predict_ds,
            train_transform=train_transform,
            val_transform=val_transform,
            test_transform=test_transform,
            predict_transform=predict_transform,
            data_fetcher=data_fetcher,
            preprocess=preprocess,
            val_split=val_split,
            batch_size=batch_size,
            num_workers=num_workers,
            sampler=sampler,
            **preprocess_kwargs,
        )
```

Fixes #778

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/flash/core/data/data_source.py
+++ b/flash/core/data/data_source.py
@@ -356,46 +356,50 @@ class DataSourceCollection(dict):
     def __init__(self, data_sources: Dict[str, DataSource], default_data_source: str):
         super().__init__(**data_sources)
         self.default_data_source = default_data_source
+        self[None] = self[default_data_source]
+        self["default"] = self[default_data_source]
 
-        for key, data_source in self.items():
-            func_name = f"from_{key.name.lower()}"
-            setattr(self, func_name, getattr(data_source, func_name, None))
+    def _execute(self, data_source_enum: DefaultDataSources, *args, **kwargs):
+        if data_source_enum in self:
+            func_name = f"from_{data_source_enum.name.lower()}"
+            return getattr(self[data_source_enum], func_name)(*args, **kwargs)
+        raise NotImplementedError
 
     def from_folders(self, *args, **kwargs):
-        raise NotImplementedError
+        return self._execute(DefaultDataSources.FOLDERS, *args, **kwargs)
 
     def from_files(self, *args, **kwargs):
-        raise NotImplementedError
+        return self._execute(DefaultDataSources.FILES, *args, **kwargs)
 
     def from_numpy(self, *args, **kwargs):
-        raise NotImplementedError
+        return self._execute(DefaultDataSources.NUMPY, *args, **kwargs)
 
     def from_tensors(self, *args, **kwargs):
-        raise NotImplementedError
+        return self._execute(DefaultDataSources.TENSORS, *args, **kwargs)
 
     def from_csv(self, *args, **kwargs):
-        raise NotImplementedError
+        return self._execute(DefaultDataSources.CSV, *args, **kwargs)
 
     def from_json(self, *args, **kwargs):
-        raise NotImplementedError
+        return self._execute(DefaultDataSources.JSON, *args, **kwargs)
 
     def from_datasets(self, *args, **kwargs):
-        raise NotImplementedError
+        return self._execute(DefaultDataSources.DATASETS, *args, **kwargs)
 
     def from_fiftyone(self, *args, **kwargs):
-        raise NotImplementedError
+        return self._execute(DefaultDataSources.FIFTYONE, *args, **kwargs)
 
     def from_dataframe(self, *args, **kwargs):
-        raise NotImplementedError
+        return self._execute(DefaultDataSources.DATAFRAME, *args, **kwargs)
 
     def from_lists(self, *args, **kwargs):
-        raise NotImplementedError
+        return self._execute(DefaultDataSources.LISTS, *args, **kwargs)
 
     def from_sentences(self, *args, **kwargs):
-        raise NotImplementedError
+        return self._execute(DefaultDataSources.SENTENCES, *args, **kwargs)
 
     def from_labelstudio(self, *args, **kwargs):
-        raise NotImplementedError
+        return self._execute(DefaultDataSources.LABELSTUDIO, *args, **kwargs)
 
 
 SEQUENCE_DATA_TYPE = TypeVar("SEQUENCE_DATA_TYPE")
--- a/flash/core/data/data_source.py
+++ b/flash/core/data/data_source.py
@@ -160,7 +160,7 @@ class DefaultDataSources(LightningEnum):
     JSON = "json"
     DATASETS = "datasets"
     FIFTYONE = "fiftyone"
-    DATAFRAME = "data_frame"
+    DATAFRAME = "dataframe"
     LISTS = "lists"
     SENTENCES = "sentences"
     LABELSTUDIO = "labelstudio"
@@ -350,6 +350,52 @@ class DataSource(Generic[DATA_TYPE], Properties, Module):
                 dataset = IterableAutoDataset(data, self, running_stage)
             dataset.__dict__.update(mock_dataset.metadata)
             return dataset
+
+
+class DataSourceCollection(dict):
+    def __init__(self, data_sources: Dict[str, DataSource], default_data_source: str):
+        super().__init__(**data_sources)
+        self.default_data_source = default_data_source
+
+        for key, data_source in self.items():
+            func_name = f"from_{key.name.lower()}"
+            setattr(self, func_name, getattr(data_source, func_name, None))
+
+    def from_folders(self, *args, **kwargs):
+        raise NotImplementedError
+
+    def from_files(self, *args, **kwargs):
+        raise NotImplementedError
+
+    def from_numpy(self, *args, **kwargs):
+        raise NotImplementedError
+
+    def from_tensors(self, *args, **kwargs):
+        raise NotImplementedError
+
+    def from_csv(self, *args, **kwargs):
+        raise NotImplementedError
+
+    def from_json(self, *args, **kwargs):
+        raise NotImplementedError
+
+    def from_datasets(self, *args, **kwargs):
+        raise NotImplementedError
+
+    def from_fiftyone(self, *args, **kwargs):
+        raise NotImplementedError
+
+    def from_dataframe(self, *args, **kwargs):
+        raise NotImplementedError
+
+    def from_lists(self, *args, **kwargs):
+        raise NotImplementedError
+
+    def from_sentences(self, *args, **kwargs):
+        raise NotImplementedError
+
+    def from_labelstudio(self, *args, **kwargs):
+        raise NotImplementedError
 
 
 SEQUENCE_DATA_TYPE = TypeVar("SEQUENCE_DATA_TYPE")
--- a/flash/image/classification/data.py
+++ b/flash/image/classification/data.py
@@ -22,7 +22,12 @@ from torch.utils.data.sampler import Sampler
 from flash.core.data.base_viz import BaseVisualization  # for viz
 from flash.core.data.callback import BaseDataFetcher
 from flash.core.data.data_module import DataModule
-from flash.core.data.data_source import DefaultDataKeys, DefaultDataSources, LoaderDataFrameDataSource
+from flash.core.data.data_source import (
+    DataSourceCollection,
+    DefaultDataKeys,
+    DefaultDataSources,
+    LoaderDataFrameDataSource,
+)
 from flash.core.data.process import Deserializer, Preprocess
 from flash.core.integrations.labelstudio.data_source import LabelStudioImageClassificationDataSource
 from flash.core.utilities.imports import _MATPLOTLIB_AVAILABLE, Image, requires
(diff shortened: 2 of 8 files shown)


flash/core/data/data_source.py
```diff
@@ -160,7 +160,7 @@ class DefaultDataSources(LightningEnum):
     JSON = "json"
     DATASETS = "datasets"
     FIFTYONE = "fiftyone"
-    DATAFRAME = "data_frame"
+    DATAFRAME = "dataframe"
```
dlangerm, line 163: unrelated?

flash/image/classification/data.py
```diff
@@ -53,6 +58,65 @@ def load_sample(self, sample: Dict[str, Any], dataset: Optional[Any] = None) ->
...
+            data_source_kwargs: Additional keyword arguments to use when constructing the data source.
+
+        Returns:
+            The constructed data module.
+        """
+        return cls(**data_source_kwargs).to_datasets(
+            (train_file, input_field, target_fields, train_images_root, train_resolver),
+            (val_file, input_field, target_fields, val_images_root, val_resolver),
+            (test_file, input_field, target_fields, test_images_root, test_resolver),
+            (predict_file, input_field, target_fields, predict_images_root, predict_resolver),
```
dlangerm, line 117: I just get the sense there has to be a better way to do this, the number of arguments to these DataSource/Module functions is pretty huge and doesn't give a very good developer experience

flash/image/classification/data.py
```diff
@@ -380,3 +444,236 @@ def show_post_tensor_transform(self, samples: List[Any], running_stage: RunningS
...
+        super().__init__(
+            data_sources={
+                DefaultDataSources.FIFTYONE: ImageFiftyOneDataSource(**data_source_kwargs),
+                DefaultDataSources.FILES: ImagePathsDataSource(),
+                DefaultDataSources.FOLDERS: ImagePathsDataSource(),
+                DefaultDataSources.NUMPY: ImageNumpyDataSource(),
+                DefaultDataSources.TENSORS: ImageTensorDataSource(),
+                DefaultDataSources.DATAFRAME: ImageClassificationDataFrameDataSource(),
+                DefaultDataSources.CSV: ImageClassificationDataFrameDataSource(),
+                DefaultDataSources.LABELSTUDIO: LabelStudioImageClassificationDataSource(),
```
dlangerm, line 460: How come none of the other data sources are passed the `kwargs`?

flash/image/classification/data.py
```diff
@@ -380,3 +444,236 @@ def show_post_tensor_transform(self, samples: List[Any], running_stage: RunningS
...
+            },
+            default_data_source=DefaultDataSources.FILES,
+        )
+
+
+class ImageClassificationDataV2(DataModule):
```
dlangerm, line 466: The naming of these modules is also a source of confusion and doesn't make it clear that they inherit from DataModule. The other classes in this library follow the convention of having the superclass name suffix the derived class. There is an ImageClassificationDataSource and an ImageClassificationPreproces but not an ImageClassificationDataModule.

flash/image/classification/data.py
```diff
@@ -380,3 +444,236 @@ def show_post_tensor_transform(self, samples: List[Any], running_stage: RunningS
...
+
+    preprocess_cls = ImageClassificationPreprocess
+    data_source_collection_cls = ImageClassificationDataSourceCollection
+
+    @classmethod
+    def from_data_source(
+        cls,
+        data_source,
```
dlangerm, line 474: So if this is to be a refactor I need to point something out about this interface (applies to the other `from_xyz` functions in the data module.

One of the most confusing things for me is the fact that this method exposes so much of the lower-level interface that it isn't clear where the arguments are supposed to come from or which ones work with which (yes the docs specify, but the interface should communicate this).

If I am sourcing data from a data source, how am I supposed to have a dataset for each fold already? Alternatively, if I have the datasets already, why do I need this extra argument for a `val_split`. 

Additionally, nothing about this interface indicates to me as a developer what I need a data fetcher for. Preprocessing makes sense here, but having it in the same argument group as the transforms does not unless you're overriding what the preprocessor has, in which case you should do that before passing it into this function, not make DataModule do it. 

One design paradigm to consider is that this should not all be one functional interface. It should be broken out into simpler functions that have argument groups that make sense with each other, or at the very least there could be a separate interface stub class that adapts to these types of function calls appropriately with arguments that make sense with each other. The developer experience in working with data pipelines would be the better for it.

flash/image/classification/data.py
```diff
@@ -380,3 +444,236 @@ def show_post_tensor_transform(self, samples: List[Any], running_stage: RunningS
...
+            preprocess=preprocess,
+            data_fetcher=data_fetcher,
+            val_split=val_split,
+            batch_size=batch_size,
+            num_workers=num_workers,
+            sampler=sampler,
+        )
+
+    @classmethod
+    def from_csv(
```
dlangerm, line 565: See the above comment about these functions with so many parameters, some of which don't make sense with each other.

flash/image/classification/data.py
```diff
@@ -380,3 +444,236 @@ def show_post_tensor_transform(self, samples: List[Any], running_stage: RunningS
...
+            )
+        """
+
+        preprocess = preprocess or cls.preprocess_cls(
+            train_transform,
+            val_transform,
+            test_transform,
+            predict_transform,
+            **preprocess_kwargs,
+        )
```
dlangerm, line 548: Again, passing in transforms as raw dictionaries results in a confusing interface for this function and unwieldy code. Now the data module has to know about how to make a preprocessor out of transforms. While handy, it adds complexity to the interface that doesn't need to be there.

flash/core/data/data_source.py
```diff
@@ -352,6 +352,52 @@ def generate_dataset(
             return dataset
 
 
+class DataSourceCollection(dict):
```
dlangerm, line 355: This is a decent idea but it adds complexity that doesn't need to be there.

DataModule already has these functions so why do they need to be here too?

dlangerm (commented):
Overall I think this method might unfortunately be adding complexity rather than removing it. The developer experience still isn't great when deciding how to integrate a preprocessing module with the rest of the data pipeline. Partially due to functions with incompatible and long argument lists, and partially due to hidden interactions between classes. Adding a data source collection may be the right call, but it doesn't help the interface and adds complexity rather than removing it.

flash/core/data/data_source.py
tchaton: Yes, agree.

flash/image/classification/data.py
tchaton: I am curious, what is your idea ?

flash/image/classification/data.py
tchaton: They don't requires it right now, but yes they should all receive it. Definitely a good point !

flash/image/classification/data.py
tchaton: I totally agree, we should make it clear with `ImageClassificationDataModule`.

flash/image/classification/data.py
tchaton: Lot of great points here !

Yes, this was designed in the following way to minimise the number of code lines at the top level and make the class_method from the datamodule as entire entry point for the data processing part. 

We made the opinionated choice that users should provide their metadata for each fold already (or they will be ignored), as Lightning supports only a single DataModule at the time.

The `val_split` argument is an example of an additional feature leaking into all `from_xyx` . This could be done in a different way, but this would break the 1 function execution which adds additional steps for the users.

 If you have ideas, don't hesitate to share some pseudo code or open PoC PRs to discuss ideas :)

tchaton:
> Overall I think this method might unfortunately be adding complexity rather than removing it. The developer experience still isn't great when deciding how to integrate a preprocessing module with the rest of the data pipeline. Partially due to functions with incompatible and long argument lists, and partially due to hidden interactions between classes. Adding a data source collection may be the right call, but it doesn't help the interface and adds complexity rather than removing it.

Yes, I think the DataSourceCollection is a good first step toward cleaner internals by separating concerns. It might open the possibly to actually separate and make the DataModule composable.

Thanks for reviewing the PoC :)

Best,
T.C

flash/image/classification/data.py
dlangerm: Well at least at first glance, there are a lot of repeated arguments (input field, target fields) and train/test/val/predict are repeated all over the place in this codebase. 

A module abstracting this series could be created so if there was a function like this where train/test/val are all in a series that module could be used to look up arguments. 

```
class FlashArgGroup:

flashArgGroup.predict.input_field to get a specific value from predict/test/train, would work like a namespace from argparse
flashArgGroup.test # to get all arguments as kwargs for the testing stage
```

Various function interfaces could be designed around that paradigm and all these functions that need a ton of arguments could just handle it with that interface.

That's just a first thought though, I'm sure there are simpler ways.

flash/image/classification/data.py
dlangerm: Check out my comments above about a `FlashArgumentGroup` class that could manage single functions like this but provide a better dev interface for grouping and repeating arguments.

PR #817, opened 2021-09-29
[Refactor] Introduce composable components

tchaton:
## What does this PR do?

Fixes # (issue)

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/flash/core/data/data_source.py
+++ b/flash/core/data/data_source.py
@@ -356,46 +356,50 @@ class DataSourceCollection(dict):
     def __init__(self, data_sources: Dict[str, DataSource], default_data_source: str):
         super().__init__(**data_sources)
         self.default_data_source = default_data_source
+        self[None] = self[default_data_source]
+        self["default"] = self[default_data_source]
 
-        for key, data_source in self.items():
-            func_name = f"from_{key.name.lower()}"
-            setattr(self, func_name, getattr(data_source, func_name, None))
+    def _execute(self, data_source_enum: DefaultDataSources, *args, **kwargs):
+        if data_source_enum in self:
+            func_name = f"from_{data_source_enum.name.lower()}"
+            return getattr(self[data_source_enum], func_name)(*args, **kwargs)
+        raise NotImplementedError
 
     def from_folders(self, *args, **kwargs):
-        raise NotImplementedError
+        return self._execute(DefaultDataSources.FOLDERS, *args, **kwargs)
 
     def from_files(self, *args, **kwargs):
-        raise NotImplementedError
+        return self._execute(DefaultDataSources.FILES, *args, **kwargs)
 
     def from_numpy(self, *args, **kwargs):
-        raise NotImplementedError
+        return self._execute(DefaultDataSources.NUMPY, *args, **kwargs)
 
     def from_tensors(self, *args, **kwargs):
-        raise NotImplementedError
+        return self._execute(DefaultDataSources.TENSORS, *args, **kwargs)
 
     def from_csv(self, *args, **kwargs):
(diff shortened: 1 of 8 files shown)


flash/core/data/process.py
```diff
@@ -413,15 +400,6 @@ def per_batch_transform_on_device(self, batch: Any) -> Any:
...
-    def available_data_sources(self) -> Sequence[str]:
-        """Get the list of available data source names for use with this
-        :class:`~flash.core.data.process.Preprocess`.
-
-        Returns:
-            The list of data source names.
-        """
-        return list(self._data_sources.keys())
-
     def data_source_of_name(self, data_source_name: str) -> DataSource:
```
dlangerm, line 403: This function still needs to go, a Registry should help with that. We'd probably need to add a deprecation notice though.

flash/core/data/data_source.py
```diff
@@ -352,6 +352,65 @@ def generate_dataset(
...
+    def from_fiftyone(self, *args, **kwargs):
+        self._execute(DefaultDataSources.FIFTYONE, *args, **kwargs)
+
+    def from_dataframe(self, *args, **kwargs):
+        self._execute(DefaultDataSources.DATAFRAME, *args, **kwargs)
+
+    def from_lists(self, *args, **kwargs):
+        self._execute(DefaultDataSources.LISTS, *args, **kwargs)
+
+    def from_sentences(self, *args, **kwargs):
```
dlangerm, line 407: Not a fan of having application-specific code here, a Registry style implementation might work better and reduce code surface. All these functions are basically just calling the same function with a different key, which sounds like a lookup to me rather than hardcoding all of these.

flash/image/classification/adapters.py
```diff
@@ -327,7 +327,7 @@ def _sanetize_batch_size(self, batch_size: int) -> int:
             warning_cache.warn(
                 "When using a meta-learning training_strategy, the batch_size should be set to 1. "
                 "HINT: You can modify the `meta_batch_size` to 100 for example by doing "
-                f"{type(self._task.task)}" + "(training_strategies_kwargs={'meta_batch_size': 100})"
+                "(training_strategies_kwargs={'meta_batch_size': 100})"
```
dlangerm, line 330: unrelated?

tests/image/classification/test_data.py
```diff
@@ -611,3 +618,46 @@ def mixup(batch, alpha=1.0):
...
+
+    assert loader._train_ds
+    assert not loader._val_ds
+    assert not loader._test_ds
+    assert loader._predict_ds
+
+    dm = ImageClassificationDataModule.from_components(
+        loader=loader,
+        preprocess=preprocess,
+    )
```
dlangerm, line 647: this interface is nice

dlangerm (commented):
Still needs work to remove all the hardcoded `from_xxx` stuff, but I like where this is going

flash/core/data/data_source.py
tchaton: Yes, I know it is not ideal, but this is primarly for IDE resolution, so contributors can follow the function call and understand where the code goes.

dlangerm: Yeah that's entirely fair, it just doesn't seem very scalable because now wouldn't we have to implement a custom "from" call both here and in the data source itself? For instance, I have a "from_config_file" method on a derived DataSource class which parses a yaml and loads up my data module.

I'm thinking it might be preferable in the longrun to abstract these away into a more scalable API? I can imaging tons of different "from_*" calls ending up making a lot of sense for different use cases all with different logic.

PR #830, opened 2021-10-04
Unified and Customizable Finetuning callback using hooks and registry.

### LINKED ISSUE #753: [RFC] Enable users of Lightning Flash to customize the finetuning callback using Hooks
## 🚀 Feature

Update the Finetuning callback to work with two hooks:
- One to get parameters to "freeze_before_training" 
- Other to get the "strategy, parameters that are to be acted upon" for the "finetune_function".

Both the hooks can be optional hooks which lets the user decide how to use the freezing and unfreezing of model parameters before and during training.

### Motivation

We currently don't let the user have a free way of fine-tuning models. For example from the QuestionAnswering task, the Fine tuning callback provides an implementation of `freeze_before_training` to freeze all the parameters of the transformer model. But there is no option to use that with a custom `finetune_function`  that the user can implement.

If the user adds a strategy like `freeze_unfreeze` or `unfreeze_milestones`, the pre-existing freeing code provided by us will be overridden.

I believe we can we support the use of hooks so that users can use the API to just provide the data (parameters / strategy and parameters) though hooks which can let them use the implementation we provide for certain methods and also have them customize others.

#### Would love to hear your thoughts on this.

karthikrangasai:
## What does this PR do?

Fixes #753 

Currently users can't use the existing `modules_to_freeze` implementation that few tasks provide and have to end up writing it back themselves. This PR provides a way for users to be able to do that.

Addition of a `FineTuningHook` class which provides a hook, 
- `modules_to_freeze` that can be implemented once in the task definition

After this, the user only needs to decide the strategy they want to use for `finetune`-ing the model with and the API for this has been made simpler:

```python
# No Freeze Strategy:
trainer.finetune(model, datamodule=datamodule, strategy="no_freeze")

# Freeze Strategy:
trainer.finetune(model, datamodule=datamodule, strategy="freeze")

# Freeze Unfreeze Strategy - requires epoch number to unfreeze the bacbone:
trainer.finetune(model, datamodule=datamodule, strategy=("freeze_unfreeze", 10))

# UnFreeze Milestones Strategy - Needs epoch numbers and number of layers to unfreeze at once:
trainer.finetune(model, datamodule=datamodule, strategy=("unfreeze_milestones", ((5, 10), 15)))
```

For users who require more command over their Finetuning implementation, can custom implement by sub-classing `BaseFinetuning`  class and providing it as the input for the strategy parameter.

## Before submitting

## PR review

## Did you have fun?


karthikrangasai:
Hello all,

3 unittests, 2 text and 1 audio , are failing because the `load_from_checkpoint` method loading weights from the cloud is looking for classes that have been removed in this PR. Hence those model weights have to updated.

Otherwise, all tests are passing fine. Looking forward to your thoughts on the PR and how to improve it.

docs/source/general/finetuning.rst
```diff
@@ -203,6 +194,49 @@ Under the hood, the pseudocode looks like:
...
+
+    datamodule = QuestionAnsweringData.from_squad_v2(
+        train_file="./data/squad_tiny/train.json",
+        val_file="./data/squad_tiny/val.json",
+    )
+
+
+    class MyCustomQuestionAnsweringTask(QuestionAnsweringTask):
+        def __init__(self, **kwargs):
+            super().__init__(**kwargs)
```
ethanwharris, line 229: This can be removed:
```suggestion
```

docs/source/general/finetuning.rst
```diff
@@ -203,6 +194,49 @@ Under the hood, the pseudocode looks like:
...
+
+custom
+------
+This strategy allows you to implement a custom finetuning logic using the freezing logic provided by Flash out of the box.
+The previosly mentioned strategies either unfreezes the complete backbone or just a fixed fraction of the backbone.
+This stratgey provides you much more control on that logic during finetuning.
+
+Here's an example where:
+
+* a distilbert for question answering backbone starts frozen
```
ethanwharris, line 207:
```suggestion
* a distilbert for question answering backbone starts frozen (the default behaviour)
```

docs/source/general/finetuning.rst
```diff
@@ -203,6 +194,49 @@ Under the hood, the pseudocode looks like:
...
+* at epoch 2 just the transformer unfreezes
```
ethanwharris, line 208:
```suggestion
* we override the ``finetune_backbone`` hook so that at epoch 2 just the transformer unfreezes
```

docs/source/general/finetuning.rst
```diff
@@ -203,6 +194,49 @@ Under the hood, the pseudocode looks like:
...
+        val_file="./data/squad_tiny/val.json",
+    )
+
+
+    class MyCustomQuestionAnsweringTask(QuestionAnsweringTask):
+        def __init__(self, **kwargs):
+            super().__init__(**kwargs)
+
+        def finetune_backbone(self, epoch: int, optimizer: Optimizer, opt_idx: int) -> None:
+            self.model.distilbert.transformer.requires_grad_(False)
```
ethanwharris, line 232: Should this be:
```suggestion
            if epoch == 2:
                self.model.distilbert.transformer.requires_grad_(False)
```

docs/source/general/finetuning.rst
```diff
@@ -203,6 +194,49 @@ Under the hood, the pseudocode looks like:
...
+        train_file="./data/squad_tiny/train.json",
+        val_file="./data/squad_tiny/val.json",
+    )
+
+
+    class MyCustomQuestionAnsweringTask(QuestionAnsweringTask):
+        def __init__(self, **kwargs):
+            super().__init__(**kwargs)
+
+        def finetune_backbone(self, epoch: int, optimizer: Optimizer, opt_idx: int) -> None:
```
ethanwharris, line 231: I think it might be cleaner for this hook to stay on the finetuning callback. That way this class which implements only custom finetuning logic would not be a new task, just a different strategy to use with the same task (and then the callback should be passed to the strategy argument).

flash/core/finetuning.py
```diff
@@ -11,147 +11,169 @@
...
 
-    def finetune_function(
-        self,
-        pl_module: LightningModule,
-        epoch: int,
-        optimizer: Optimizer,
-        opt_idx: int,
-    ) -> None:
-        pass
+class FinetuningStrategies(LightningEnum):
```
ethanwharris, line 25: I like the Enum :smiley: I think it would be neat to also keep the separate class for each strategy even if they just override init with a hard coded strategy key, that way it can be backwards compatible.

ethanwharris (commented):
Awesome work, the ``get_backbone_for_finetuning`` hook is really neat. I have some reservations about where the ``finetune_backbone`` hook should live :smiley:

docs/source/general/finetuning.rst
karthikrangasai: The idea behind `finetune_backbone` is that if we implement it in the `finetune_function` of the `FlashBaseFinetuning` class by creating a new custom callback, we don't get to use the `freeze_before_training` logic provided by Flash.

flash/core/finetuning.py
karthikrangasai: So, should I revert the changes and map each Enum member to a class implementation ?

ethanwharris: No need to revert, could just have e.g.:
```
class FreezeUnfreeze(FlashBaseFinetuning):
    def __init__(
        self,
        strategy_metadata: Optional[Union[int, Tuple[int, int]]] = None,
        train_bn: bool = True,
    ):
        super().__init__(FinetuningStrategies.FREEZE_UNFREEZE, strategy_metadata, train_bn)
```

docs/source/general/finetuning.rst
ethanwharris: What about if we had the current finetune_backbone hook just as part of the finetune callback rather than the model? That way the finetune callbacks could be used outside of flash and just implement the finetune_backbone hook to add custom unfreezing behaviour?

ethanwharris: Also, should the hook maybe be called something like unfreeze_at_epoch(epoch: int, ...) to make the intended usage (unfreezing some weights) more clear?

karthikrangasai: We can make the custom mode a callable based on.

```
def my_unfreezing_logic(pl_module, epoch, optimizer, opt_idx):
    <perform the logic>
trainer.finetune(model, datamodule, strategy=my_unfreezing_logic)
```

So then we don't need the `custom` key and directly infer that it's a custom function using existing freezing logic.

flash/core/finetuning.py
karthikrangasai: Got it.

docs/source/general/finetuning.rst
tchaton: I would prefer not to have this on the task personally. Not all models have a backbone. Or we should make it more general. `on_finetune_start`

flash/core/finetuning.py
```diff
@@ -11,147 +11,169 @@
...
         opt_idx: int,
-    ) -> None:
-        pass
+        strategy_metadata: int,
+    ):
+        unfreeze_epoch: int = strategy_metadata
+        if epoch != unfreeze_epoch:
+            return
 
+        get_backbone_for_finetuning = getattr(pl_module, "get_backbone_for_finetuning", None)
```
tchaton, line 90: IMO, we should avoid the following API. This is quite restrictive and won't scale well to more complex tasks, especially when we start multi modals.

tchaton: I would prefer an advanced scheduler as this one: https://github.com/PyTorchLightning/pytorch-lightning/pull/10198. Which can be used to control the fine-tuning strategy.

karthikrangasai: I understand and agree, this approach might become restrictive for further expansion.

I will currently keep the API simple with just registry based routing of the finetuning callbacks with the added first hook.

docs/source/api/core.rst
```diff
@@ -44,9 +44,7 @@ _____________________
     :template: classtemplate.rst
 
     ~flash.core.finetuning.FlashBaseFinetuning
-    ~flash.core.finetuning.FreezeUnfreeze
-    ~flash.core.finetuning.NoFreeze
-    ~flash.core.finetuning.UnfreezeMilestones
```
ethanwharris, line 49: I guess these can be added back

docs/source/general/finetuning.rst
```diff
@@ -127,21 +127,13 @@ Every finetune strategy can also be customized.
 
 freeze_unfreeze
 ---------------
-By default, in this strategy the backbone is frozen for 5 epochs then unfrozen:
+The freeze_unfreeze strategy keeps the backbone frozen until a certain epoch (provided through the input) after which the backbone unfrozen.
```
ethanwharris, line 130: Maybe something like:
```suggestion
The ``freeze_unfreeze`` strategy keeps the backbone frozen until a certain epoch (provided in a tuple to the ``strategy`` argument) after which the backbone will be unfrozen.
```

flash/core/finetuning.py
```diff
@@ -11,147 +11,187 @@
...
+    def __init__(
+        self,
+        strategy_key: FinetuningStrategies,
+        strategy_metadata: Optional[Union[int, Tuple[Tuple[int, int], int]]] = None,
+        train_bn: bool = True,
+    ):
         """
         Args:
+            strategy_key: The finetuning strategy to be used. (Check :meth"`Trainer.finetune` for the available
+             strategies.)
```
ethanwharris, line 51:
```suggestion
            strategy_key: The finetuning strategy to be used. See :meth:`~flash.core.trainer.Trainer.finetune`
                for the available strategies.
```

flash/core/model.py
```diff
@@ -583,9 +587,55 @@ def configure_optimizers(self) -> Union[Optimizer, Tuple[List[Optimizer], List[_
...
+                )
+            if strategy[0] == "unfreeze_milestones" and not (
+                isinstance(strategy[1], Tuple)
+                and isinstance(strategy[1][0], Tuple)
+                and isinstance(strategy[1][1], int)
+                and isinstance(strategy[1][0][0], int)
+                and isinstance(strategy[1][0][1], int)
+            ):
+                raise MisconfigurationException(
+                    "`unfreeze_milestones` stratgey only accepts the format Tuple[Tuple[int, int], int]. HINT example: "
```
ethanwharris, line 625:
```suggestion
                    "`unfreeze_milestones` strategy only accepts the format Tuple[Tuple[int, int], int]. HINT example: "
```

flash/core/trainer.py
```diff
@@ -175,48 +175,37 @@ def finetune(
...
-                Default strategies can be enabled with these strings:
+                Default strategies can be enabled with these inputs:
 
-                - ``"no_freeze"``,
-                - ``"freeze"``,
-                - ``"freeze_unfreeze"``,
-                - ``"unfreeze_milestones"``.
+                - ``"no_freeze"``
+                - ``"freeze"``
+                - ``("freeze_unfreeze", integer)``
```
ethanwharris, line 185:
```suggestion
                - ``("freeze_unfreeze", integer: unfreeze_epoch)``
```

flash/core/trainer.py
```diff
@@ -175,48 +175,37 @@ def finetune(
...
+                - ``("unfreeze_milestones", ((integer, integer), integer))``
```
ethanwharris, line 186:
```suggestion
                - ``("unfreeze_milestones", ((integer: unfreeze_epoch_num_layers, integer: unfreeze_epoch_all_layers), integer: num_layers))``
```

ethanwharris (approved):
Awesome, really like the direction this is going, LGTM :smiley: Some small comments

flash/core/model.py
```diff
@@ -583,9 +587,55 @@ def configure_optimizers(self) -> Union[Optimizer, Tuple[List[Optimizer], List[_
...
+                raise MisconfigurationException(
+                    f"First input of `strategy` in a tuple configuration should be a string within"
+                    f" {_DEFAULTS_FINETUNE_STRATEGIES[3:]}"
+                )
+            if strategy[0] == "freeze_unfreeze" and not isinstance(strategy[1], int):
+                raise MisconfigurationException(
+                    "`freeze_unfreeze` stratgey only accepts one integer denoting the epoch number to switch."
+                )
+            if strategy[0] == "unfreeze_milestones" and not (
+                isinstance(strategy[1], Tuple)
```
tchaton, line 618: Could we move the validator to the callback directly?

karthikrangasai: This was initially present in the `trainer.finetune` method just to throw errors before the finetuning starts.
But I moved it to the `Task.configure_finetune_callback` method to maintain the semantics of the code.
Wouldn't pushing to specific callbacks throw the Exceptions a little late in the pipeline ?

flash/core/model.py
```diff
@@ -583,9 +587,38 @@ def configure_optimizers(self) -> Union[Optimizer, Tuple[List[Optimizer], List[_
...
+            if not isinstance(strategy[0], str) or strategy[0] not in self.available_finetuning_strategies():
+                raise MisconfigurationException(
+                    f"First input of `strategy` in a tuple configuration should be a string within"
+                    f" {_DEFAULTS_FINETUNE_STRATEGIES[3:]}"
+                )
+            finetuning_strategy_fn: Callable = self.finetuning_strategies.get(key=strategy[0])
+            finetuning_strategy_metadata = {"strategy_metadata": strategy[1], "train_bn": train_bn}
+        else:
+            raise MisconfigurationException(
+                "strategy should be a ``pytorch_lightning.callbacks.BaseFinetuning``"
```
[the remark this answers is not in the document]
karthikrangasai: strategy is the name of the argument. Hence the casing. Should I change it to `strategy` ?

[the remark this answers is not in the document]
ethanwharris: Ah, yeah, that would be good :smiley:

flash/core/finetuning.py
```diff
@@ -11,147 +11,201 @@
...
 
+    def freeze_before_training(self, pl_module: Union[Module, Iterable[Union[Module, Iterable]]]) -> None:
+        if self.strategy != FinetuningStrategies.NO_FREEZE:
+            modules_to_freeze = getattr(pl_module, "modules_to_freeze", None)
+            if modules_to_freeze is None:
+                raise AttributeError(
+                    "LightningModule missing instance method 'modules_to_freeze'."
+                    "Please, implement the method which returns NoneType or a Module or an Iterable of Modules."
+                )
+            modules = modules_to_freeze()
```
ethanwharris, line 89:
```suggestion
            modules = self._get_modules_to_freeze(pl_module=pl_module)
```

Issue #760, opened 2021-09-14
Add TextEmbedder Task


SeanNaren:
## 🚀 Feature
Using https://github.com/UKPLab/sentence-transformers create a Flash Task for embedding sentences. Sentence Transformers also supports fine-tuning the model onto your own set of data, to improve semantic embedding representations.

### Motivation

It's common to require text search, and sentence-transformers provides great results + an API to fine-tune/predict to get sentence representations out. It could also be paired with [FAISS](https://github.com/facebookresearch/faiss) really simply to provide fast/accurate search results across large datasets.

abhijithneilabraham:
@SeanNaren can I take this up?

SeanNaren:
hey @abhijithneilabraham absolutely! please do :) more than happy to assist if you need help!

abhijithneilabraham:
How do you propose I do this? Open a new module  named `embeddings` inside https://github.com/PyTorchLightning/lightning-flash/tree/master/flash/text and return sentence embeddings?

SeanNaren:
Yep exactly! It would also be possible to fine-tune the model as well on new data, which would be awesome

PR #838, opened 2021-10-06
Sentence embedder

abhijithneilabraham:
First step to #760 

Created a SentenceEmbedder task that would return the embeddings from Sentence Transformers. Also created template for API for training using Sentence Transformers api.
```
from flash.text.embeddings import SentenceEmbedder

sentence_embedder=SentenceEmbedder()
print(sentence_embedder.generate_embeddings(["Today was a good day"]))
```

@SeanNaren @tchaton


SeanNaren:
This is looking much much nicer @abhijithneilabraham! Definitely a great place for us to start :)

To really get this fully featured I suggest defining an example script that uses the flash trainer to fine-tune, test and predict, like so: https://github.com/PyTorchLightning/lightning-flash/blob/master/flash_examples/text_classification.py

Once you manage to figure out a dataset, I can help you to get the dataset onto the flash S3 bucket to auto-download :)

PR #945, opened 2021-11-08
pin TV version

Borda:
## What does this PR do?

need to pin TorchVision version as some internal dependencies broke as TV moved/renamed `utils`

Fixes:
```
    from torchvision.models.utils import load_state_dict_from_url
ModuleNotFoundError: No module named 'torchvision.models.utils'
```

## Before submitting

## PR review

## Did you have fun?


justusschock (changes requested):
I haven't found this anywhere in our codebase. Also this is just an import from torch that shifted. So with the following we should be good to go without this restriction:

```
try:
    from torch.hub import load_state_dict_from_url  # noqa: 401
except ImportError:
    from torch.utils.model_zoo import load_url as load_state_dict_from_url  # noqa: 401
```

Borda:
> I haven't found this anywhere in our codebase. Also this is just an import from torch that shifted.

no it is not in our codebase, but in our dependencies, IceVision >> TIMM

justusschock:
then maybe we should just open an issue there? :)

Borda:
@ethanwharris thoughts?

ethanwharris (changes requested):
@Borda I'm not sure we shoyuld do this, all of our tests are passing so it's not clear what this is solving. Also, since each torchvision version is pinned to a specific torch version this would be equivalent to pinning `torch < 1.10` which is quite extreme

PR #953, opened 2021-11-11
Add new input classes in `input_base.py`

### LINKED ISSUE #923: Add new `Input` object outside of `InputTransform`
This will be a big one. See #871 for a PoC

Has the following aims:
- Massively reduce the complexity of data loading, psuedo code for the `from_*` method would be:
```
# create the input object (e.g. `ImageFolderInput`)
# load the datasets for each stage from the input object
# instantiate the datamodule with the data sets
```
- Implement the class methods directly on the datamodule class rather than on the base class. This enables us to then document exactly how to use the method for that specific task rather than just a generic docstring. In addition, we can use the required Input object directly rather than from a directory or registry. That way a user can follow the code straight to the actual code that is loading their data in a couple of clicks.

Some challenges:
- This will need to work with Flash zero, which means we still need a way to know what classmethods are supported. This could maybe be done in a follow-up.
- This will cause a lot of code changes, one idea is to apply the change in one PR per data-type (e.g. image tasks first then text, etc.)

ethanwharris:
## What does this PR do?

Fixes #923

Note for the reviewer, this is a transition PR from the previous API with the AutoDataset to the new one without.

## Before submitting

## PR review

## Did you have fun?


flash/graph/classification/data.py
```diff
@@ -60,6 +61,33 @@ class GraphClassificationData(DataModule):
...
+        test_dataset: Optional[Dataset] = None,
+        predict_dataset: Optional[Dataset] = None,
+        train_transform: Optional[Dict[str, Callable]] = None,
+        val_transform: Optional[Dict[str, Callable]] = None,
+        test_transform: Optional[Dict[str, Callable]] = None,
+        predict_transform: Optional[Dict[str, Callable]] = None,
+        **data_module_kwargs,
+    ) -> "GraphClassificationData":
+        return cls(
+            GraphDatasetInput(RunningStage.TRAINING, train_dataset),
```
tchaton, line 78: Neat ! I like this :) Very clear, no magic anymore.

flash/video/classification/data.py
```diff
@@ -388,3 +348,220 @@ class VideoClassificationData(DataModule):
     """Data module for Video classification tasks."""
 
     input_transform_cls = VideoClassificationInputTransform
+
+    @classmethod
+    def from_files(
+        cls,
+        train_files: Optional[Sequence[str]] = None,
```
tchaton, line 355: Should add `input_cls = VideoClassificationFiftyOneInput` there ? Users could modify the dataset class but also, we won't need a registry. We could check all the `from_*` signatures and extract the default class from it associated to the function. Just an idea.

flash/video/classification/data.py
```diff
@@ -388,3 +348,220 @@ class VideoClassificationData(DataModule):
...
+        decode_audio: bool = False,
+        decoder: str = "pyav",
+        **data_module_kwargs,
+    ) -> "VideoClassificationData":
+        return cls(
+            VideoClassificationFilesInput(
+                RunningStage.TRAINING,
+                train_files,
+                train_targets,
+                clip_sampler=clip_sampler,
```
tchaton, line 379: Let's use dataset_kwargs to remove some lines.

flash/video/classification/data.py
```diff
@@ -191,76 +189,54 @@ def _make_encoded_video_dataset(self, data) -> "LabeledVideoDataset":
...
-        labeled_video_paths = LabeledVideoPaths(list(zip(filepaths, targets)))
-
-        ds: LabeledVideoDataset = LabeledVideoDataset(
-            labeled_video_paths,
-            self.clip_sampler,
-            video_sampler=self.video_sampler,
-            decode_audio=self.decode_audio,
-            decoder=self.decoder,
+
+        dataset = LabeledVideoDataset(
```
tchaton, line 232: Can we share this code? There is a lot of duplicated code.

ananyahjha93 (approved):

flash/core/data/io/input_base.py
```diff
@@ -0,0 +1,173 @@
...
+def _has_len(data: Union[Sequence, Iterable]) -> bool:
+    try:
+        len(data)
+        return True
+    except (TypeError, NotImplementedError):
+        return False
+
+
+def _validate_input(input: "InputBase"):
+    if input.data is not None:
```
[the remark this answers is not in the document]
ethanwharris: None is considered a valid value since we use it to determine whether the `Input` is truthy or falsey which is then used by the datamodule to know whether to provide a dataset / dataloader for that stage.

flash/video/classification/data.py
ethanwharris: Not easily at the moment, but I think a further refactor / cleanup of the video data sources will be in order once the data pipeline V2 work is finished.

flash/video/classification/data.py
ethanwharris: Could do, worth exploring when we get to that stage :smiley:

tchaton (approved):
LGTM !

PR #970, opened 2021-11-15
Fix RTD build

ethanwharris:
## What does this PR do?

The read the docs build environment (where the "READTHEDOCS" environment variable is set to True) doesn't include torch so it mocks the `IterableDataset` import. Whatever type that mock is doesn't then work as a base for a metaclass. So we instead replace the import with `object` since `type(object) = type` and `type` is the natural base for a Python metaclass.

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/flash/core/data/io/input_base.py
+++ b/flash/core/data/io/input_base.py
@@ -13,9 +13,9 @@
 # limitations under the License.
 import functools
 import sys
-from typing import Any, cast, Dict, Iterable, MutableMapping, Sequence, Tuple, Union
+from typing import Any, cast, Dict, Iterable, MutableMapping, Sequence, Tuple, TYPE_CHECKING, Union
 
-from torch.utils.data import Dataset, IterableDataset
+from torch.utils.data import Dataset
 
 from flash.core.data.properties import Properties
 from flash.core.utilities.stages import RunningStage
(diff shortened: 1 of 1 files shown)


flash/core/data/io/input_base.py
```diff
@@ -26,6 +27,12 @@
     GenericMeta = type
 
 
+if not os.environ.get("READTHEDOCS", False):
+    from torch.utils.data import IterableDataset
+else:
+    IterableDataset = object
```
awaelchli, line 33: This is the strangest thing ever. If we look at line 94:

```py
class _IterableInputMeta(_InputMeta, type(IterableDataset)):
```

which means if the if condition above holds, we will get
```py
type(IterableDataset) = torch.utils.data._typing._DataPipeMeta
```
and if it does not hold we will get
```py
type(object) = type
```

Just mentioning it in case this was not the intention.

ethanwharris: This is intentional :smiley: The read the docs build environment doesn't include torch so it mocks the `IterableDataset` import. Whatever type that mock is doesn't then work as a base for a metaclass. So we instead replace the import with object since `type(object) = type` and `type` is the natural base for a Python metaclass. I've added a comment to the code to make this more clear :smiley:

awaelchli (approved):

PR #990, opened 2021-11-22
add predict_kwargs in ObjectDetectionModel in order to filter the pre…

### LINKED ISSUE #954: Allow to specify the detection_threshold in the prediction method of ObjectDetector
## 🚀 Feature
Allow in the method predict of the ObjectDetector to specify the parameter detection_threshold that is used to filter the bounding box in base of the confidence of the model

### Motivation

In Icevision is possible to specify the detection_threshold that is used to filter the bounding box predicted by the model. If it's not specified the value is set to default at 0.5. Instead using ObjectDetector is not allow to specify this parameter and then the prediction are always filtered using the default threshold

### Pitch

Allow to specify a custom value for the threshold used to filter the bounding boxes

### Additional context

This is a screenshot of the predict method used in Icevision. If you agree that is can be useful to add this functionality I can work on it and open a PR
Schermata 2021-11-12 alle 12 58 43

Actis92:
…diction using custom threshold and manage other parameters

## What does this PR do?

In this PR I have added the possibility to specify predict_kwargs for ObjectDetector that can be used to add additional parameters that are used during the prediction step. For example is possible to specify a custom threshold in order to filter the bounding boxes predicted by Icevision. And added setter method in order to allow user to update the values
Fixes #954


ethanwharris (commented):
Looking good :smiley: @Actis92 mind also adding this to the instance segmentation and keypoint detection tasks? They are also using the IceVision adapter under the hood :smiley:

Actis92:
@ethanwharris I have added this parameter to the instance segmentation and keypoint detection tasks. Let me know if it's needed something more :)

flash/image/detection/model.py
```diff
@@ -75,3 +79,7 @@ def __init__(
     def _ci_benchmark_fn(self, history: List[Dict[str, Any]]) -> None:
         """This function is used only for debugging usage with CI."""
         # todo
+
+    def set_predict_kwargs(self, value):
+        """This function is used to update the kwargs used for the prediction step."""
+        self.adapter.predict_kwargs = value
```
ethanwharris, line 85: I think it would be cleaner to do this with properties. Something like:
```suggestion
    @property
    def predict_kwargs(self) -> Dict[str, Any]:
        """The kwargs used for the prediction step."""
        return self.adapter.predict_kwargs
        
    @predict_kwargs.setter
    def predict_kwargs(predict_kwargs: Dict[str, Any]):
        self.adapter.predict_kwargs = predict_kwargs
```

One day we should find a way for this sort of thing to exist on the adapter and automatically be attached here...

ethanwharris (commented):
Awesome, I think we could make the setter slightly cleaner but just a minor point :smiley: Also, mind updating the CHANGELOG?

flash/image/detection/model.py
Actis92: Hi @ethanwharris I will do the changes, both CHANGELOG and the properties. For what concern how to do it automatically, I think that a way is that to automatically update the attribute __dict__ of the class, taking the properties from the adapter or something like that :)

Actis92: I have done the changes

ethanwharris (approved):
Nice! LGTM :smiley:

PR #1013, opened 2021-12-01
Data Pipeline V2: Refactor Deserializer into Serve Input

### LINKED ISSUE #916: Refactor `Deserializer` to be a `DataSource` / `Input`
The deserializers should just be viewed as a `ServeInput`, perhaps with an empty `load_data` and the addition of the sample input property that our sanity checking uses.

tchaton:
## What does this PR do?

Fixes #916

This change is temporary until we get merge the ServeInput and add `serve_load_sample` directly to the Input base class.

## Before submitting

## PR review

## Did you have fun?


flash/core/data/io/input_base.py
```diff
@@ -232,3 +233,26 @@ def __iter__(self):
...
+    ) -> None:
+        if hasattr(self, "serve_load_data"):
+            raise MisconfigurationException("`serve_load_data` shouldn't be implemented.")
+
+        super().__init__(RunningStage.SERVING, data_pipeline_state=data_pipeline_state)
+
+    def serve_load_sample(self, sample: Any) -> List[Any]:
+        raise NotImplementedError
+
+    def deserialize(self, sample: Any) -> Any:
```
ethanwharris, line 251: Should we instead have this use `__call__`?

flash/core/model.py
```diff
@@ -809,7 +809,13 @@ def build_data_pipeline(
         if deserializer is None or type(deserializer) is Deserializer:
             deserializer = getattr(input_transform, "deserializer", deserializer)
 
-        data_pipeline = DataPipeline(input, input_transform, output_transform, deserializer, output)
+        data_pipeline = DataPipeline(
+            input=input,
+            input_transform=input_transform,
+            output_transform=output_transform,
+            deserializer=deserializer,
```
ethanwharris, line 816: Might be good to refer to it as `serve_input` or something instead of `deserializer`

flash/image/data.py
```diff
@@ -60,7 +60,7 @@ def image_loader(filepath: str):
 
 class ImageDeserializer(Deserializer):
```
ethanwharris, line 61: Maybe this naming would be better:
```suggestion
class ImageServeInput(ServeInput):
```

flash/core/data/process.py
```diff
@@ -12,30 +12,14 @@
...
-    @property
-    @abstractmethod
-    def example_input(self) -> str:
-        raise NotImplementedError
-
-    def __call__(self, sample: Any) -> Any:
-        return self.deserialize(sample)
 
 
 class DeserializerMapping(Deserializer):
```
ethanwharris, line 25: We should probably delete `DeserializerMapping`, we don't have a use case for it right now, can always bring it back later if needed

ethanwharris (commented):
Looks good :smiley: I guess the existing deserializers should be renamed to end in `ServeInput` and will also need to extend serve input rather than deserializer

flash/image/data.py
tchaton: Let's do this when we get rid of the Deserializer type entirely.

flash/core/data/io/input_base.py
tchaton: Done !

flash/core/model.py
tchaton: I think we would do this with the DataPipeline API change.

flash/core/data/process.py
tchaton: This would be done in another PR

Issue #964, opened 2021-11-12
Onboard text inputs to new Input object


ethanwharris:
Tracking issue #955

PR #1022, opened 2021-12-03
Onboard text classification inputs to new object

ethanwharris:
## What does this PR do?

Part of #964 

## Before submitting

## PR review

## Did you have fun?


flash/text/classification/data.py
```diff
@@ -11,247 +11,168 @@
...
-    def __init__(self, backbone: str, max_length: int, use_fast: bool = True, **kwargs):
+    def __init__(self, max_length: int):
         super().__init__()
-        self.backbone = backbone
-        self.tokenizer = AutoTokenizer.from_pretrained(backbone, use_fast=use_fast, **kwargs)
         self.max_length = max_length
 
+    @property
+    @lru_cache(maxsize=None)
+    def tokenizer(self):
```
tchaton, line 50: Neat !

flash/text/classification/data.py
```diff
@@ -11,247 +11,168 @@
...
-        return ex
 
+class TextClassificationInput(ClassificationInput):
     @staticmethod
-    def _multilabel_target(targets: List[str], element: Dict[str, Any]) -> Dict[str, Any]:
-        targets = [element.pop(target) for target in targets]
-        element[DataKeys.TARGET] = targets
+    def _resolve_target(target_keys: Union[str, List[str]], element: Dict[str, Any]) -> Dict[str, Any]:
+        if not isinstance(target_keys, List):
+            element[DataKeys.TARGET] = element.pop(target_keys)
```
tchaton, line 66: Would this work as expected ?

flash/text/classification/data.py
```diff
@@ -11,247 +11,168 @@
...
-                    hf_dataset = hf_dataset.map(partial(self._transform_label, label_to_class_mapping, target))
-
-                # rename label column
-                hf_dataset = hf_dataset.rename_column(target, DataKeys.TARGET)
+            hf_dataset = hf_dataset.map(partial(self._resolve_target, target_keys))
+            targets = hf_dataset.to_dict()[DataKeys.TARGET]
+            self.load_target_metadata(targets)
+
+            # If we had binary multi-class targets then we also know the labels (column names)
+            if self.target_mode is TargetMode.MULTI_BINARY and isinstance(target_keys, List):
```
tchaton, line 88: Why do you override the state there ? If the state is shared among train, val. This would override the train one.

flash/text/classification/data.py
```diff
@@ -11,247 +11,168 @@
...
-        return (hf_dataset, *other)
+class TextClassificationCSVInput(TextClassificationInput):
+    def load_data(
+        self,
+        csv_file: PATH_TYPE,
+        input_key: str,
+        target_keys: Optional[Union[str, List[str]]] = None,
+        max_length: int = 128,
+    ) -> Dataset:
+        dataset_dict = load_dataset("csv", data_files={"train": str(csv_file)})
```
tchaton, line 125: Would this work if we put a dummy key ? train might be confusing there.

flash/text/classification/model.py
```diff
@@ -30,6 +32,13 @@
     from transformers.modeling_outputs import Seq2SeqSequenceClassifierOutput, SequenceClassifierOutput
 
 
+@dataclass(unsafe_hash=True, frozen=True)
+class TextClassificationBackboneState(ProcessState):
+    """The ``TextClassificationBackboneState`` records the ``backbone`` in use by the ``TextClassifier``."""
+
+    backbone: str
```
tchaton, line 39: Do we want to expend the state to anything behind passed along ?

flash/text/seq2seq/core/data.py
```diff
@@ -23,16 +23,41 @@
...
     from datasets import DatasetDict, load_dataset
     from transformers import AutoTokenizer, default_data_collator
 
 
+class Seq2SeqDeserializer(Deserializer):
+    @requires("text")
+    def __init__(self, backbone: str, max_length: int, use_fast: bool = True, **kwargs):
+        super().__init__()
+        self.backbone = backbone
+        self.tokenizer = AutoTokenizer.from_pretrained(backbone, use_fast=use_fast, **kwargs)
```
tchaton, line 41: Do you want to use the property too ?

tchaton (approved):
Looks good, some comments.

flash/text/seq2seq/core/data.py
ethanwharris: Will do when we get to Seq2Seq tasks in a follow up :smiley:

flash/text/classification/data.py
ethanwharris: The classification state is only ever shared from the train data (becuase labels are always inferred from the train data rather than val. or test). In this case, the labels are binary, so we assume that a one in a column corresponds to having that label. So the labels inferred by our classification utils are `None` but we replace them with the target keys from the dataset.

flash/text/classification/data.py
ethanwharris: This is just a preprocessing step before our own target handling mechanism. So it doesn't matter what comes out here as we will later handle it properly with our classification utilities.

flash/text/classification/data.py
ethanwharris: Yes, good catch :smiley: Actually, it looks like we don't need a key at all haha

PR #1029, opened 2021-12-06
Data Pipeline Refactor: Improve new DataModule

tchaton:
## What does this PR do?

Fixes # (issue)

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/flash/core/data/input_transform.py
+++ b/flash/core/data/input_transform.py
@@ -998,6 +998,9 @@ def _make_collates(input_transform: "InputTransform", on_device: bool, collate: 
 def _create_collate_input_transform_processors(
     input_transform: "InputTransform", callbacks: List[FlashCallback]
 ) -> Tuple[_InputTransformProcessor, _InputTransformProcessor]:
+    """This utility is used to create the 2 `_InputTransformProcessor` objects contain the transforms used as the
+    DataLoader `collate_fn` and the `on_after_batch_transfer` hook."""
+
     from flash.core.data.data_pipeline import DataPipeline
 
     prefix: str = _STAGES_PREFIX[input_transform.running_stage]
--- a/flash/core/data/input_transform.py
+++ b/flash/core/data/input_transform.py
@@ -116,16 +116,6 @@ class InputTransform(Properties):
             "transform": self._transform,
         }
 
-    @property
-    def dataloader_collate_fn(self):
-        """Generate the function to be injected within the DataLoader as the collate_fn."""
-        return self._create_collate_input_transform_processors()[0]
-
-    @property
-    def on_after_batch_transfer_fn(self):
-        """Generate the function to be injected after the on_after_batch_transfer from the LightningModule."""
-        return self._create_collate_input_transform_processors()[1]
-
     ########################
     # PER SAMPLE TRANSFORM #
     ########################
@@ -936,64 +926,6 @@ class InputTransform(Properties):
 
     def __getitem__(self, placement: InputTransformPlacement) -> Callable:
         return self._transform[placement]
-
-    def _make_collates(self, on_device: bool, collate: Callable) -> Tuple[Callable, Callable]:
-        if on_device:
-            return self._identity, collate
-        return collate, self._identity
-
-    def _create_collate_input_transform_processors(self) -> Tuple[Any]:
-        from flash.core.data.data_pipeline import DataPipeline
-
-        prefix: str = _STAGES_PREFIX[self.running_stage]
-
-        per_batch_transform_overridden: bool = DataPipeline._is_overridden_recursive(
-            "per_batch_transform", self, InputTransform, prefix=prefix
-        )
-
-        per_sample_transform_on_device_overridden: bool = DataPipeline._is_overridden_recursive(
-            "per_sample_transform_on_device", self, InputTransform, prefix=prefix
-        )
-
-        is_per_overridden = per_batch_transform_overridden and per_sample_transform_on_device_overridden
-        if self._collate_in_worker_from_transform is None and is_per_overridden:
-            raise MisconfigurationException(
-                f"{self.__class__.__name__}: `per_batch_transform` and `per_sample_transform_on_device` "
-                f"are mutually exclusive for stage {self.running_stage}"
-            )
-
-        if isinstance(self._collate_in_worker_from_transform, bool):
-            worker_collate_fn, device_collate_fn = self._make_collates(
-                not self._collate_in_worker_from_transform, self._collate
-            )
-        else:
-            worker_collate_fn, device_collate_fn = self._make_collates(
-                per_sample_transform_on_device_overridden, self._collate
-            )
-
-        worker_collate_fn = (
-            worker_collate_fn.collate_fn
-            if isinstance(worker_collate_fn, _InputTransformProcessor)
-            else worker_collate_fn
-        )
-
-        worker_input_transform_processor = _InputTransformProcessor(
-            self,
-            worker_collate_fn,
-            self._per_sample_transform,
-            self._per_batch_transform,
-            self.running_stage,
-        )
-        device_input_transform_processor = _InputTransformProcessor(
-            self,
-            device_collate_fn,
-            self._per_sample_transform_on_device,
-            self._per_batch_transform_on_device,
-            self.running_stage,
-            apply_per_sample_transform=device_collate_fn != self._identity,
-            on_device=True,
-        )
-        return worker_input_transform_processor, device_input_transform_processor
 
 
 @dataclass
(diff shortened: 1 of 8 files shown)


flash/core/data/input_transform.py
```diff
@@ -1051,3 +987,68 @@ def create_transform(
...
+def _make_collates(input_transform: "InputTransform", on_device: bool, collate: Callable) -> Tuple[Callable, Callable]:
+    if on_device:
+        return input_transform._identity, collate
+    return collate, input_transform._identity
+
+
+def _create_collate_input_transform_processors(
+    input_transform: "InputTransform", callbacks: List[FlashCallback]
+) -> Tuple[_InputTransformProcessor, _InputTransformProcessor]:
+    """This utility is used to create the 2 `_InputTransformProcessor` objects contain the transforms used as the
```
ethanwharris, line 1001:
```suggestion
    """This utility is used to create the 2 `_InputTransformProcessor` objects which contain the transforms used as the
```

flash/core/data/new_data_module.py
```diff
@@ -58,14 +85,19 @@ class DataModule(DataModule):
...
-        test_dataset: Optional[Input] = None,
-        predict_dataset: Optional[Input] = None,
+        train_dataset: Optional[Dataset] = None,
+        val_dataset: Optional[Dataset] = None,
+        test_dataset: Optional[Dataset] = None,
+        predict_dataset: Optional[Dataset] = None,
+        train_transform: Optional[INPUT_TRANSFORM_TYPE] = None,
+        val_transform: Optional[INPUT_TRANSFORM_TYPE] = None,
+        test_transform: Optional[INPUT_TRANSFORM_TYPE] = None,
+        predict_transform: Optional[INPUT_TRANSFORM_TYPE] = None,
```
ethanwharris, line 100: This is quite confusing to me. Why do we need two ways to provide transforms?

flash/core/data/io/input_base.py
```diff
@@ -236,18 +236,6 @@ def register_input_transform(
...
-        if self.transform:
-            self.transform.running_stage = self.running_stage
-            return self.transform.dataloader_collate_fn
-
-    @property
-    def on_after_batch_transfer_fn(self) -> Optional[Callable]:
-        if self.transform:
-            self.transform.running_stage = self.running_stage
-            return self.transform.on_after_batch_transfer_fn
-
```
ethanwharris, line 250: Not sure about this change. It makes sense to me that the input would contain everything related to the transforms as that is the standard contract of a pytorch dataset. We are just extending it to include a collate and a device transform.

flash/core/data/new_data_module.py
```diff
@@ -314,23 +405,92 @@ def _create_input(
...
+                :class:`~flash.core.data.data_module.DataModule`.
+            input_transform: The :class:`~flash.core.data.io.input_transform.InputTransform` to pass to the
+                :class:`~flash.core.data.data_module.DataModule`. If ``None``, ``cls.input_transform_cls``
+                will be constructed and used.
+            val_split: The ``val_split`` argument to pass to the :class:`~flash.core.data.data_module.DataModule`.
+            batch_size: The ``batch_size`` argument to pass to the :class:`~flash.core.data.data_module.DataModule`.
+            num_workers: The ``num_workers`` argument to pass to the :class:`~flash.core.data.data_module.DataModule`.
+            sampler: The ``sampler`` to use for the ``train_dataloader``.
+            input_transform_kwargs: Additional keyword arguments to use when constructing the input_transform.
+                Will only be used if ``input_transform = None``.
```
ethanwharris, line 470: I think these can be removed.

flash/core/data/new_data_module.py
```diff
@@ -314,23 +405,92 @@ def _create_input(
...
-        if cls.flash_datasets_registry is None:
-            raise MisconfigurationException("The class attribute `flash_datasets_registry` should be set. ")
-        cls.flash_datasets_registry(fn=input_cls, name=enum)
+    def register_input(cls, enum: Union[str, LightningEnum], input_cls: Type[Input]) -> None:
+        if cls.inputs_registry is None:
+            raise MisconfigurationException("The class attribute `inputs_registry` should be set. ")
+        cls.inputs_registry(fn=input_cls, name=enum)
+
+    @classmethod
+    def from_datasets(
```
ethanwharris, line 428: Since we have `from_datasets` which will wrap any torch Dataset in an input object and then call the constructure, why do we need to also accept `Dataset` in the `__init__`?

ethanwharris (commented):
Not sure about some of the changes here. It seems like adding transforms as optional arguments to the datamodule is a step back in readability and I'm not sure what it adds. If a user wants to use their own dataset then they would use `from_datasets` which would wrap their datasets and transforms in an input object anyway, so all transform logic could still exist within the Input class

flash/core/data/new_data_module.py
tchaton: Hey @ethanwharris,

I believe the core question is: Should we support dataset being pyre pytorch dataset?

If the answer is no, then we can remove those transforms.
Otherwise, we would need to support the transform in order to simplify onboarding.

IMO, I would prefer to support pure PyTorch dataset there and `from_datasets` transform the sample into dictionaries which might not be the user desired behaviour.

tchaton: And if a user provides both an input with transform and a transform, it would raise a MisConfigurationExecption, so there is still 1 way to do something:

Use Input -> Transform are owned by the Input
Use PyTorch Dataset -> You can pass InputTransform as a utility method.

flash/core/data/io/input_base.py
tchaton: Added back !

flash/core/data/new_data_module.py
tchaton: Ok, you could do this now:

```py
    dm = DataModule.from_datasets(
        train_dataset=RandomDataset(64, 32),
        val_dataset=RandomDataset(64, 32),
        test_dataset=RandomDataset(64, 32),
        train_transform=TrainInputTransform,
        val_transform=_add_hundred,
        input_cls=Input,
        batch_size=3,
    )
    batch = next(iter(dm.train_dataloader()))
    assert batch[0][0] == 2
    batch = next(iter(dm.val_dataloader()))
    assert batch[0][0] == 101
    batch = next(iter(dm.test_dataloader()))
    assert batch[0][0] == 1
```

Let me remove support for it then.

ethanwharris: > Otherwise, we would need to support the transform in order to simplify onboarding.

With a standard PyTorch Dataset you would usually apply your transforms inside the `__getitem__`. So users would need to change their code here in order to provide the transforms this way.

> and from_datasets transform the sample into dictionaries which might not be the user desired behaviour.

Do you have an example of a case where that is undesired? All of our tasks require dictionaries so this seems like quite a big step back in simplicity in order to support a way of loading data that wouldn't work with any of our tasks.

Personally, I would rather remove the automatic conversion of samples to a dictionary in the DatasetInput. That way if a user wants to bring their own data set from PyTorch they would always be responsible for formatting their samples correctly and can have non-dictionary samples if that is what they want.

tchaton: Done !

flash/core/data/new_data_module.py
tchaton: Done !

flash/core/data/new_data_module.py
tchaton: Done !

flash/core/data/io/input_transform.py
```diff
@@ -522,10 +522,11 @@ def __init__(
         stage: RunningStage,
         apply_per_sample_transform: bool = True,
         on_device: bool = False,
+        callbacks: List[FlashCallback] = None,
```
ethanwharris, line 525:
```suggestion
        callbacks: Optional[List[FlashCallback]] = None,
```

flash/core/data/new_data_module.py
```diff
@@ -43,6 +61,10 @@ class DataModule(DataModule):
         val_dataset: Dataset for validating model performance during training. Defaults to None.
         test_dataset: Dataset to test model performance. Defaults to None.
         predict_dataset: Dataset for predicting. Defaults to None.
+        train_transform: Either a callable or a :class:`~flash.core.data.input_transform.InputTransform` transform.
+        val_transform: Either a callable or a :class:`~flash.core.data.input_transform.InputTransform` transform.
+        test_transform: Either a callable or a :class:`~flash.core.data.input_transform.InputTransform` transform.
+        predict_transform: Either a callable or a :class:`~flash.core.data.input_transform.InputTransform` transform.
```
ethanwharris, line 67:
```suggestion
```

ethanwharris (approved):
LGTM :smiley:

flash/core/data/io/input_transform.py
tchaton: Done !

flash/core/data/new_data_module.py
tchaton: Done !

PR #1110, opened 2022-01-12
update doc object detection

### LINKED ISSUE #1107: Object detection example not working for custom transformation Flash version 0.6.0
Here is described how use the ObjectDetector Module, and how pass custom transformations. As you can see in the example it is suggested to define a dictionary, but using it with Flash 0.6.0 it didn't work.
But i obtain the following error:
File "/Users/lucaactisgrosso/PycharmProjects/lit-tracking/venv/lib/python3.7/site-packages/flash/core/data/io/input_transform.py", line 1112, in create_transform
    raise MisconfigurationException(f"The format for the transform isn't correct. Found {transform}")
pytorch_lightning.utilities.exceptions.MisconfigurationException: The format for the transform isn't correct. Found {'per_sample_transform': IceVisionTransformAdapter()}

because in the function create_transform in the file flash/core/data/io/input_transform.py not works with dictionary as input value. Someone knows what is the correct way to pass custom transformation with the latest release of Flash?

Actis92:
Fixes #1107

In this PR I have updated the documentation for Object Detection as we have discussed in the issue #1107

### DIFF
--- a/docs/source/reference/object_detection.rst
+++ b/docs/source/reference/object_detection.rst
@@ -93,6 +93,7 @@ creating a subclass of :class:`~flash.core.data.io.input_transform.InputTransfor
     from flash.core.integrations.icevision.transforms import IceVisionTransformAdapter
     from flash.image import ObjectDetectionData
 
+
     @dataclass
     class BrightnessContrastTransform(InputTransform):
         image_size: int = 128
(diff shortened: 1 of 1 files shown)


ethanwharris (approved):
Awesome! Thanks for this @Actis92, LGTM :smiley:

PR #1152, opened 2022-02-02
Add `FlashRegistry` of Available Heads for `flash.image.ImageClassifier`

### LINKED ISSUE #1150: Create `FlashRegistry` of Available Heads for `flash.image.ImageClassifier`
Hey folks! This is my first time opening an issue here. Great project!

## ❓ Questions and Help

After reading the documentation as well as source code, I find that `flash.image.ImageClassifier` has no flash registry for available heads. There's one for all other `flash.image` tasks but not the classification task.

Right now, the only way I see to add a heads registry is to create a new class, inherit from `flash.image.ImageClassifier`, copy and update the `__init__()` method to fetch heads from a registry. 

Is there an easy way to do this? Although it's possible to create a `nn.module` for the head and pass it to `flash.image.ImageClassifier`, I would like to keep a registry and simply pass the head id and additional kwargs, just like one would do for loading the model backbone.

AjinkyaIndulkar:
## What does this PR do?
This PR adds a `FlashRegistry` to `flash.image.ImageClassifier`. 

Fixes #1150

### Added:

- ✨ added `IMAGE_CLASSIFIER_HEADS` flash registry for `flash.image.ImageClassifier`
- ✅ added unit test to test new heads registry

### Updated:

- ⚡ updated `flash.image.ImageClassifier` to use `IMAGE_CLASSIFIER_HEADS` flash registry

## Notes:

- The scope of implementation is limited to the heads currently available for `ImageClassifier`.
- All unit tests have passed locally (with `make test`)

### DIFF
--- a/flash/image/classification/heads.py
+++ b/flash/image/classification/heads.py
@@ -27,7 +27,7 @@ def _load_linear_head(num_features: int, num_classes: int) -> nn.Module:
     Args:
         num_features: Number of input features.
         num_classes: Number of output classes.
-        
+
     Returns:
         nn.Module: Linear head.
     """
--- a/flash/image/classification/heads.py
+++ b/flash/image/classification/heads.py
@@ -25,8 +25,9 @@ def _load_linear_head(num_features: int, num_classes: int) -> nn.Module:
     """Loads a linear head.
 
     Args:
-        num_features (int): Number of input features.
-        num_classes (int): Number of output classes.
+        num_features: Number of input features.
+        num_classes: Number of output classes.
+        
     Returns:
         nn.Module: Linear head.
     """
@@ -36,6 +37,4 @@ def _load_linear_head(num_features: int, num_classes: int) -> nn.Module:
 IMAGE_CLASSIFIER_HEADS(
     partial(_load_linear_head),
     name="linear",
-    namespace="image/classification",
-    providers="torch",
 )
--- a/flash/image/classification/heads.py
+++ b/flash/image/classification/heads.py
@@ -13,7 +13,6 @@
 # limitations under the License.
 from functools import partial
 
-import torch
 from torch import nn
 
 from flash.core.registry import FlashRegistry  # noqa: F401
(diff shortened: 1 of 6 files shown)


flash/image/classification/heads.py
```diff
@@ -0,0 +1,66 @@
...
+    """Loads a linear head.
+
+    Args:
+        num_features (int): Number of input features.
+        num_classes (int): Number of output classes.
+
+    Returns:
+        nn.Module: Linear head.
+    """
+    return LinearHead(num_features, num_classes)
```
ethanwharris, line 58: We could simplify a bit here:
```suggestion
def _load_linear_head(num_features: int, num_classes: int) -> nn.Module:
    """Loads a linear head.

    Args:
        num_features (int): Number of input features.
        num_classes (int): Number of output classes.

    Returns:
        nn.Module: Linear head.
    """
    return nn.Linear(num_features, num_classes)
```

AjinkyaIndulkar:
> @AjinkyaIndulkar This looks great! Left a small comment, could you also add a line to `CHANGELOG.md`?

Sure. I'll add your suggestion as well.

flash/image/classification/heads.py
AjinkyaIndulkar: resolved in 0e2cb6d.

flash/image/classification/heads.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+def _load_linear_head(num_features: int, num_classes: int) -> nn.Module:
+    """Loads a linear head.
+
+    Args:
+        num_features (int): Number of input features.
+        num_classes (int): Number of output classes.
+    Returns:
+        nn.Module: Linear head.
```
ethanwharris, line 31:
```suggestion

    Args:
        num_features: Number of input features.
        num_classes: Number of output classes.
        
    Returns:
        nn.Module: Linear head.
```

flash/image/classification/heads.py
```diff
@@ -0,0 +1,41 @@
...
+    """
+    return nn.Linear(num_features, num_classes)
+
+
+IMAGE_CLASSIFIER_HEADS(
+    partial(_load_linear_head),
+    name="linear",
+    namespace="image/classification",
+    providers="torch",
```
ethanwharris, line 40: Can get rid of these two I think:
```suggestion
```

ethanwharris (approved):
Awesome, LGTM :smiley:

PR #1318, opened 2022-04-28
simplify torch imports - Tensor & Module

Borda:
## What does this PR do?

Notably, following 2 changes in the imports are made:

- `from torch import Tensor`
- `from torch import nn` and using `nn.ModuleDict, nn.Module, nn.ModuleList`

instead of: `import torch` and using:

- `torch.Tensor`
- `torch.nn.Module, torch.nn.ModuleDict, torch.nn.ModuleList`

## Before submitting

## PR review

## Did you have fun?


flash/core/data/utils.py
```diff
@@ -18,9 +18,9 @@
 from typing import Any, Callable, Dict, Iterable, Mapping, Optional, Set
 
 import requests
-import torch
 import urllib3
 from pytorch_lightning.utilities.apply_func import apply_to_collection
+from torch.nn import Module, ModuleDict, ModuleList
```
krshrimali, line 23: Thanks for the refactoring, @Borda - but I'm not very sure if this is a great idea. And this might be just me, I usually prefer having a namespace around such modules where it's not immediately clear where they come from. So something like `nn.ModuleDict` is more clear than `ModuleDict` and doing `from torch.nn import ModuleDict`.

krshrimali (approved):
Thanks, @Borda for the refactoring! I'm a little skeptical about the changes, in some places, but overall it looks fine. I'll wait for Ethan's comments, and then we can merge this if all is good :)

flash/core/data/utils.py
Borda: the long imports tend to cut lines also I believe import is slightly faster...

krshrimali:
Hi, @Borda - I'm okay with replacing `torch.Tensor` with `Tensor`, but hesitant on replacing other imports. And this can be just me, but I've noticed that using `nn.Module` makes it clear that it comes from the `nn` module of PyTorch, while using `Module` doesn't tell a lot where it's coming from. Same goes for `ModuleDict` and other imports. Though, I'm fine if in a separate PR later we decide to refactor / simplify the imports for `torch.Tensor`, but I still think it's fine the way the code is right now. With these simplified imports (though we can not help much about torchvision as using `T.ToTensor()` like structure is quite standard), we get these errors like `T not defined` (https://github.com/Lightning-AI/lightning-flash/issues/1403) if the user doesn't have torchvision installed. Though, this has to do more with raising better error messages if the module is not available, but just sharing the issue here.

If you think otherwise, I'm happy to discuss this further. Please feel free to re-open it in that case.

Thanks!

Borda:
> noticed that using `nn.Module` makes it clear that it comes from the `nn` module of PyTorch

well, do you know any other `Module` coming from a different library?

> I'm okay with replacing `torch.Tensor` with `Tensor`

so why close full PR and not ever the second change?

krshrimali:
> > noticed that using `nn.Module` makes it clear that it comes from the `nn` module of PyTorch
> 
> well, do you know any other `Module` coming from a different library?

As I said, it makes it clear that it's coming from `nn`. It's not about whether there are ambiguities, it's `from library import *` vs `import library` and then using `library.*` practice. While both of them are right, it's more about what is preferred, and in my message, I just said that I would prefer doing `import library` and then doing `library.*`.

> > I'm okay with replacing `torch.Tensor` with `Tensor`
> 
> so why close full PR and not ever the second change?

Cool, I'll remove the other import changes, and keep `torch.Tensor` -> `Tensor` changes, but still I would say for consistency, either we should do: `from library import *` or `import library` -> `library.*` behavior.

Open to discussions on this. Thanks!

PR #1327, opened 2022-05-04
add Trivial Aug for image classification

Borda:
## What does this PR do?

Adding a minimal transformation which shall be dataset independent

## Before submitting
- [ ] Did you read the **[contributor guideline](https://github.com/PyTorchLightning/lightning-flash/tree/master/.github/CONTRIBUTING.md)**, Pull Request section?
- [ ] Did you make sure your **PR does only one thing**, instead of bundling different changes together?

## PR review

## Did you have fun?

### DIFF
--- a/flash/core/utilities/imports.py
+++ b/flash/core/utilities/imports.py
@@ -121,7 +121,7 @@ else:
 
 if Version:
     _TORCHVISION_GREATER_EQUAL_0_9 = _compare_version("torchvision", operator.ge, "0.9.0")
-    _TORCHVISION_GREATER_EQUAL_0_11 = _compare_version("torchvision", operator.ge, "0.9.0")
+    _TORCHVISION_GREATER_EQUAL_0_11 = _compare_version("torchvision", operator.ge, "0.11.0")
     _PL_GREATER_EQUAL_1_4_3 = _compare_version("pytorch_lightning", operator.ge, "1.4.3")
     _PL_GREATER_EQUAL_1_4_0 = _compare_version("pytorch_lightning", operator.ge, "1.4.0")
     _PL_GREATER_EQUAL_1_5_0 = _compare_version("pytorch_lightning", operator.ge, "1.5.0")
--- a/flash/core/utilities/imports.py
+++ b/flash/core/utilities/imports.py
@@ -121,6 +121,7 @@ else:
 
 if Version:
     _TORCHVISION_GREATER_EQUAL_0_9 = _compare_version("torchvision", operator.ge, "0.9.0")
+    _TORCHVISION_GREATER_EQUAL_0_11 = _compare_version("torchvision", operator.ge, "0.9.0")
     _PL_GREATER_EQUAL_1_4_3 = _compare_version("pytorch_lightning", operator.ge, "1.4.3")
     _PL_GREATER_EQUAL_1_4_0 = _compare_version("pytorch_lightning", operator.ge, "1.4.0")
     _PL_GREATER_EQUAL_1_5_0 = _compare_version("pytorch_lightning", operator.ge, "1.5.0")
--- a/flash/image/classification/input_transform.py
+++ b/flash/image/classification/input_transform.py
@@ -18,7 +18,12 @@ import torch
 
 from flash.core.data.io.input_transform import InputTransform
 from flash.core.data.transforms import kornia_collate
-from flash.core.utilities.imports import _ALBUMENTATIONS_AVAILABLE, _TORCHVISION_AVAILABLE, requires
+from flash.core.utilities.imports import (
+    _ALBUMENTATIONS_AVAILABLE,
+    _TORCHVISION_AVAILABLE,
+    _TORCHVISION_GREATER_EQUAL_0_11,
+    requires,
+)
 
 if _TORCHVISION_AVAILABLE:
     from torchvision import transforms as T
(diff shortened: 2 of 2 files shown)


flash/image/classification/input_transform.py
```diff
@@ -51,7 +51,13 @@ def input_per_sample_transform(self):
 
     def train_input_per_sample_transform(self):
         return T.Compose(
-            [T.ToTensor(), T.Resize(self.image_size), T.Normalize(self.mean, self.std), T.RandomHorizontalFlip()]
+            [
+                T.ToTensor(),
+                T.Resize(self.image_size),
+                T.Normalize(self.mean, self.std),
+                T.RandomHorizontalFlip(),
```
ligaz, line 58: Do we really need the `RandomHorizontalFlip` in this case when `TrivialAugmentWide` is applied before hand?

flash/core/utilities/imports.py
```diff
@@ -121,6 +121,7 @@ class Image:
 
 if Version:
     _TORCHVISION_GREATER_EQUAL_0_9 = _compare_version("torchvision", operator.ge, "0.9.0")
+    _TORCHVISION_GREATER_EQUAL_0_11 = _compare_version("torchvision", operator.ge, "0.9.0")
```
ligaz, line 124: Typo:
```suggestion
    _TORCHVISION_GREATER_EQUAL_0_11 = _compare_version("torchvision", operator.ge, "0.11.0")
```

flash/image/classification/input_transform.py
```diff
@@ -50,9 +55,14 @@ def input_per_sample_transform(self):
...
-        return T.Compose(
-            [T.ToTensor(), T.Resize(self.image_size), T.Normalize(self.mean, self.std), T.RandomHorizontalFlip()]
-        )
+        transforms = [T.TrivialAugmentWide()] if _TORCHVISION_GREATER_EQUAL_0_11 else []
+        transforms += [
+            T.ToTensor(),
+            T.Resize(self.image_size),
+            T.RandomHorizontalFlip(),
+            T.Normalize(self.mean, self.std),
+        ]
```
ligaz, line 64: This will move the `TrivialAugmentWide` as the first transform. Is this expected? IMO it will better to apply the augmentation after the normalization.

Borda: yes, but it does not work as `TrivialAugmentWide` requires uint8 and `ToTensor` convert to float... so this is the only place it can be used...

flash/image/classification/input_transform.py
Borda: not sure what geometrical transformation is applied, check some samples and most changes were about color intensities...

flash/image/classification/input_transform.py
ligaz: We can get an inspiration from the way it is done in `torchvision`:
https://github.com/pytorch/vision/blob/e288f6ca01a5009ab32f1db010b31f14756e0303/references/classification/presets.py#L18-L41

flash/image/classification/input_transform.py
krshrimali: That's a good question, @ligaz - I was looking at the source code of this transform. Here are the transforms it applies, if I'm not wrong:

https://github.com/pytorch/vision/blob/62740807c18e68bb0acd85895dca527f9a655bd5/torchvision/prototype/transforms/_auto_augment.py#L428:L449

```python
    _AUGMENTATION_SPACE = {
        "Identity": (lambda num_bins, height, width: None, False),
        "ShearX": (lambda num_bins, height, width: torch.linspace(0.0, 0.99, num_bins), True),
        "ShearY": (lambda num_bins, height, width: torch.linspace(0.0, 0.99, num_bins), True),
        "TranslateX": (lambda num_bins, height, width: torch.linspace(0.0, 32.0, num_bins), True),
        "TranslateY": (lambda num_bins, height, width: torch.linspace(0.0, 32.0, num_bins), True),
        "Rotate": (lambda num_bins, height, width: torch.linspace(0.0, 135.0, num_bins), True),
        "Brightness": (lambda num_bins, height, width: torch.linspace(0.0, 0.99, num_bins), True),
        "Color": (lambda num_bins, height, width: torch.linspace(0.0, 0.99, num_bins), True),
        "Contrast": (lambda num_bins, height, width: torch.linspace(0.0, 0.99, num_bins), True),
        "Sharpness": (lambda num_bins, height, width: torch.linspace(0.0, 0.99, num_bins), True),
        "Posterize": (
            lambda num_bins, height, width: cast(torch.Tensor, 8 - (torch.arange(num_bins) / ((num_bins - 1) / 6)))
            .round()
            .int(),
            False,
        ),
        "Solarize": (lambda num_bins, height, width: torch.linspace(255.0, 0.0, num_bins), False),
        "AutoContrast": (lambda num_bins, height, width: None, False),
        "Equalize": (lambda num_bins, height, width: None, False),
    }
```

IMO, we could have dropped horizontal flip since this transform does have enough augmentations, _but_, since it's not available for all torchvision versions, so I'll honestly put a condition.

If this augmentation is available, use it, else include `RandomHorizontalFlip`.

```python
transforms = [
    T.ToTensor(),
    T.Resize(self.image_size),
    T.Normalize(self.mean, self.std)
]

transforms += [T.TrivialAugmentWide()] if _TORCHVISION_GREATER_EQUAL_0_11 else [T.RandomHorizontalFlip()]
```

**Note**: I think it's a better idea to convert the input to a tensor first, resize and normalize before applying this transform?Let me know what you guys think. @Borda @ligaz

krshrimali (changes requested):
Thanks, @Borda for the PR. I appreciate the thought of adding auto augmentation to Flash, however - I've raised a concern [here](https://github.com/PyTorchLightning/lightning-flash/pull/1327#discussion_r865535725). I'm not very sure if it's a great idea to _perform lots of transformations_ by default. We can think of having an argument to support auto augmentation policies later (probably can create an issue on this to start a discussion).

But yes, these are just my thoughts, and would love to hear yours! :)

flash/image/classification/input_transform.py
```diff
@@ -50,9 +55,14 @@ def input_per_sample_transform(self):
         return T.Compose([T.ToTensor(), T.Resize(self.image_size), T.Normalize(self.mean, self.std)])
 
     def train_input_per_sample_transform(self):
-        return T.Compose(
-            [T.ToTensor(), T.Resize(self.image_size), T.Normalize(self.mean, self.std), T.RandomHorizontalFlip()]
-        )
+        transforms = [T.TrivialAugmentWide()] if _TORCHVISION_GREATER_EQUAL_0_11 else []
```
ethanwharris, line 58: This seems really weird to me. It means that for different PL versions you will get different augmentations, so also potentially quite dramatically different performance...

ethanwharris (changes requested):
Not sure about this one. I think that we should try to make the experience as consistent as possible across different versions of torch

Issue #1323, opened 2022-05-03
Error while using `show_train_batch`


aniketmaurya:
## 🐛 Bug

Load the [Herbarium-2022](https://www.kaggle.com/competitions/herbarium-2022-fgvc9) dataset using `ImageClassificationData.from_data_frame` and do `dm.show_train_batch()` it throws the following error.

```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Input In [81], in <cell line: 1>()
----> 1 train_dm.show_train_batch()

File ~/miniconda3/envs/herbarium/lib/python3.10/site-packages/flash/core/data/data_module.py:520, in DataModule.show_train_batch(self, hooks_names, reset)
    518 """This function is used to visualize a batch from the train dataloader."""
    519 stage_name: str = _STAGES_PREFIX[RunningStage.TRAINING]
--> 520 self._show_batch(stage_name, hooks_names, reset=reset)

File ~/miniconda3/envs/herbarium/lib/python3.10/site-packages/flash/core/data/data_module.py:513, in DataModule._show_batch(self, stage, func_names, reset)
    511     _ = next(iter_dataloader)
    512 data_fetcher: BaseVisualization = self.data_fetcher
--> 513 data_fetcher._show(stage, func_names)
    514 if reset:
    515     self.data_fetcher.batches[stage] = {}

File ~/miniconda3/envs/herbarium/lib/python3.10/site-packages/flash/core/data/base_viz.py:103, in BaseVisualization._show(self, running_stage, func_names_list)
    102 def _show(self, running_stage: RunningStage, func_names_list: List[str]) -> None:
--> 103     self.show(self.batches[running_stage], running_stage, func_names_list)

File ~/miniconda3/envs/herbarium/lib/python3.10/site-packages/flash/core/data/base_viz.py:115, in BaseVisualization.show(self, batch, running_stage, func_names_list)
    113 hook_name = f"show_{func_name}"
    114 if _is_overridden(hook_name, self, BaseVisualization):
--> 115     getattr(self, hook_name)(batch[func_name], running_stage)

File ~/miniconda3/envs/herbarium/lib/python3.10/site-packages/flash/image/classification/data.py:1167, in MatplotlibVisualization.show_load_sample(self, samples, running_stage)
   1165 def show_load_sample(self, samples: List[Any], running_stage: RunningStage):
   1166     win_title: str = f"{running_stage} - show_load_sample"
[...]

PR #1339, opened 2022-05-10
Bugfix: `_show_images_and_labels` - flatten axes to allow plotting for subplots

krshrimali:
## What does this PR do?

Fixes https://github.com/PyTorchLightning/lightning-flash/issues/1323.

The following script will fail:

```python
import matplotlib.pyplot as plt
import numpy as np

fig, axs = plt.subplots(2, 2)
for i, ax in enumerate(axs):
    ax.imshow(np.random.randint(1, 255, (5, 5, 3)))
```

Note that the object "ax" is still a `list/ndarray` here, and not an Axes object, so we should flatten the `axs` object before looping through it. (for some discussion: https://stackoverflow.com/questions/43957776/numpy-ndarray-object-has-no-attribute-imshow)

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,7 +26,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ### Fixed
 
-- Fixed `_show_images_and_labels` for subplots with rows > 1. ([#1339](https://github.com/PyTorchLightning/lightning-flash/pull/1315))
+- Fixed image classification data `show_train_batch` for subplots with rows > 1. ([#1339](https://github.com/PyTorchLightning/lightning-flash/pull/1315))
 
 - Fixed support for all the versions (including the latest and older) of `baal`. ([#1315](https://github.com/PyTorchLightning/lightning-flash/pull/1315))
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ### Fixed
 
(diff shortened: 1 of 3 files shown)


tests/image/classification/test_data.py
```diff
@@ -221,6 +221,40 @@ def test_from_filepaths_visualise(tmpdir):
...
+@pytest.mark.skipif(not _MATPLOTLIB_AVAILABLE, reason="matplotlib isn't installed.")
+def test_from_filepaths_visualise_subplots_exceding_max_cols(tmpdir):
+    tmpdir = Path(tmpdir)
+
+    (tmpdir / "e").mkdir()
+    _rand_image().save(tmpdir / "e_1.png")
+
+    train_images = [
+        str(tmpdir / "e_1.png"),
+    ] * 8
```
krshrimali, line 234: We were not able to catch this bug because of a missing text having a number of images > max_cols * N ( N > 1 ). Here, the number of images we pass is 8, and hence the subplot will have 4 rows and 2 columns.

Also ensured that this test fails on master (without the fix in this PR).

CHANGELOG.md
```diff
@@ -26,6 +26,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ### Fixed
 
+- Fixed `_show_images_and_labels` for subplots with rows > 1. ([#1339](https://github.com/PyTorchLightning/lightning-flash/pull/1315))
```
ethanwharris, line 29: Just a slight adjustment to communicate the fix from a user perspective
```suggestion
- Fixed image classification data `show_train_batch` for subplots with rows > 1. ([#1339](https://github.com/PyTorchLightning/lightning-flash/pull/1315))
```

ethanwharris (approved):
LGTM :smiley:

PR #1387, opened 2022-07-13
Refactor file loading to use fsspec

### LINKED ISSUE #1298: Expose better API for file I/O in DataModules
## 🚀 Feature
Expose better API for file I/O handling that will allow easier usage of fsspec / external file systems.

### Motivation
We have a use case where training needs to stream data from cloud storage (S3 or GCS). Usually, we use fsspec for that. 

In the current version of `lightning-flash` we're using `ImageClassificationData.from_data_frame` where the input `DataFrame` contains a list of paths to files stored in the cloud (with `s3://` or `gcs://` prefixes). Although it's possible to override the behaviour of the `ImageClassificationFilesInput` to handle loading from external file systems, it seems like it's API was not designed to do so, especially because `flash.image.data.image_loader` cannot be easily replaced (e.g. via `kwargs`) in the `flash.image.data.ImageFilesInput.load_sample`. Right now I need to override 2 methods: `resolver` (via constructor) and `load_sample` (via inheritance), like in the example below.

### Pitch

Introduce parameters to high level API of the `*FilesInput` classes that will allow to specify actual file loading instead of relying on built-in Python's `open` and `os.path` APIs, assuming that everything is local.
This will make cloud training deployments easier.

### Alternatives
* force the user to always have full dataset locally (or on mounted volume)
* always use fsspec for file I/O across the framework (might be too large change for such simple case)
* leave the API as is, leading to more code on the project level

#### Current solution
```python
class FSSpecMixin:
    @lru_cache()
    def _get_fs(self, protocol: str):
        return fsspec.filesystem(protocol)

    def get_fs(self, protocol: str, **storage_options) -> AbstractFileSystem:
        if storage_options:
            return fsspec.filesystem(protocol, **storage_options)
        else:
            return self._get_fs(protocol)

    def get_fs_for_path(self, filepath: str, **storage_options) -> AbstractFileSystem:
        protocol, _ = get_protocol_and_path(filepath)
        return self.get_fs(protocol, **storage_options)

class ImageFSSpecInput(ImageFilesInput, FSSpecMixin):
    def _load_image_using_fspec(self, filepath: str):
        with self.get_fs_for_path(filepath).open(filepath, "rb") as f:
            return Image.open(f).convert("RGB")

    def load_sample(self, sample: Dict[str, Any]) -> Dict[str, Any]:

        # HERE YOU CANNOT PLUG-IN THE FSSPEC LOGIC WITHOUT COPYING THE CODE FROM THE PARENT CLASS
        # DUE TO THE flash.image.data.image_loader BEING "HARD-CODED"

        filepath = sample[DataKeys.INPUT]
        sample[DataKeys.INPUT] = self._load_image_using_fspec(filepath)

        sample = ImageInput.load_sample(self, sample) # <--- NOTE THE CALL TO EXPLICIT PARENT CLASS
        
        sample[DataKeys.METADATA]["filepath"] = filepath
        return sample

class ImageClassificationDataFrameInputFSSpec(
    ImageClassificationDataFrameInput, ImageFSSpecInput
):
    def load_data(
        self,
        data_frame: pd.DataFrame,
        input_key: str,
        target_keys: Optional[Union[str, List[str]]] = None,
        root: Optional[PATH_TYPE] = None,
        resolver: Optional[Callable[[Optional[PATH_TYPE], Any], PATH_TYPE]] = None,
        target_formatter: Optional[TargetFormatter] = None,
    ) -> List[Dict[str, Any]]:
        def _fspec_resolver(_: Any, filepath: str) -> str:
            if not self.get_fs_for_path(filepath).exists(filepath):
                raise ValueError(f"File {filepath} does not exist")
            return filepath

        return super().load_data(
            data_frame, input_key, target_keys, root, _fspec_resolver, target_formatter
        )
```

Usage:
```python
datamodule = ImageClassificationData.from_data_frame(input_cls=ImageClassificationDataFrameInputFSSpec, **kwargs)
```

#### Alternative nasty workaround
One can also monkey-patch the `flash.image.data.image_loader` using `from unittest.mock import patch` in a context manager, but I don't want such code in production environment.

### Additional context
If there is a better entrypoint for overriding the file I/O in DataModules, I would be glad to learn about it. Thanks!

ethanwharris:
## What does this PR do?

Refactors file loading to use fsspec and a series of other enhancements:
- added formats for audio loading (but drops mp3, couldn't get that to work)
- added support for ".tsv" files in data frame loading
- standardised all loaded spectrograms to be float32
- tests every supported format with every loader

Fixes #1298 

## Before submitting
- [ ] Did you make sure your **PR does only one thing**, instead of bundling different changes together?
- [x] Did you make sure to **update the documentation** with your changes?

## PR review

## Did you have fun?


krshrimali (commented):
Hi, @ethanwharris - wow! This looks great. 🎉 🔥 🚀 Thanks for the summary in the description as well. Sorry for the late comments, I don't really have any suggestions but a couple of questions:

1. Since URLs are supported, apart from the release notes, we need a way to tell the users that they can pass the URLs instead of downloading the CSV files using `download_data`. What are your opinions on raising a `UserWarning` or a message for calls to `download_data`? Something like:

```bash
UserWarning: For Lightning Flash v0.8.0+, URLs are now supported to be passed directly. If your use-case supports it, you can now skip the call to `download_data`, and directly pass the URL to your file.
```

2.  Curious to know what went wrong with the MP3 files? If you have any info on it, we can probably create an issue and come back to it later.

ethanwharris:
@krshrimali Thanks for the comments!
1. I don't think we should do that, `download_data` is stilll a valid thing to do just not always the simplest. We should just make sure that the option of passing a URL is clear from our docs (still some work to do there with docstrings I think).
2. It looks like MP3 support was only added in the latest version of soundfile and I couldn't get it to work. We should figure out what the requirements are for it and then it could be added back (as long as we can make it run in our CI).

PR #1396, opened 2022-07-15
[1/2] Use any task as an embedder for any layer

ethanwharris:
## What does this PR do?

This PR adds support for calling `as_embedder` on any task to convert it to an embedder. The embeddings retain gradient information so they could be used for training the model weights.

This approach uses an exception to break out of the forward in intermediate layers to avoid excess computation.

Still to come in follow-up: docs

## Before submitting
- [ ] Did you make sure to **update the documentation** with your changes? - Coming in 2/2

## PR review

## Did you have fun?


tests/core/test_model.py
```diff
@@ -290,6 +291,32 @@ def test_model_download(tmpdir, cls, filename):
...
+
+    def predict_step(self, batch, batch_idx: int, dataloader_idx: int = 0):
+        return self.model(batch)
+
+
+def test_as_embedder():
+    embedder = DummyTask().as_embedder("model.1")
+
+    assert isinstance(embedder, Embedder)
+    assert embedder.predict_step(torch.rand(10, 10), 0, 0).size(1) == 30
```
krshrimali, line 312: Minor suggestion: probably better if we don't hard-code, and use `out_features` from the respective layer:

```suggestion
def test_as_embedder():
    layer_number = "1"
    embedder = DummyTask().as_embedder("model." + layer_number)

    assert isinstance(embedder, Embedder)
    assert (
        embedder.predict_step(torch.rand(10, 10), 0, 0).size(1)
        == embedder.model.model._modules[layer_number].out_features
    )
```

tests/core/test_model.py
```diff
@@ -290,6 +291,33 @@ def test_model_download(tmpdir, cls, filename):
         assert isinstance(task, cls)
 
 
+class DummyTask(Task):
+    def __init__(self):
+        super().__init__()
+
+        self.model = nn.Sequential(
```
justusschock, line 298: very nit: for testing: Can we use another attribute name here (not `self.model`)? This would ease debugging since it is also called `self.model` inside the embedder class.

tests/core/test_model.py
```diff
@@ -290,6 +291,33 @@ def test_model_download(tmpdir, cls, filename):
...
+            nn.Linear(30, 40),
+        )
+
+    def predict_step(self, batch, batch_idx: int, dataloader_idx: int = 0):
+        return self.model(batch)
+
+
+def test_as_embedder():
+    layer_number = "1"
+    embedder = DummyTask().as_embedder(f"model.{layer_number}")
```
justusschock, line 310: just a question: Should we also allow to only call `.as_embedder(layer_number)`? I.e. splitting after the last separator '.' and check those as well? I think almost all builtin tasks only have a single model, right?

justusschock (approved):
LGTM

flash/core/utilities/embedder.py
```diff
@@ -0,0 +1,82 @@
...
+    def _register_hook(self):
+        if self._module is not None:
+            self._handle = self._module.register_forward_hook(self._hook)
+
+    def _remove_hook(self):
+        if self._handle is not None:
+            self._handle.remove()
+        self._handle = None
+
+    def training_step(self, batch: Any, batch_idx: int) -> Any:
```
otaj, line 65: Are you sure we don't to be training embedders?

otaj (approved):
LGTM!

flash/core/utilities/embedder.py
ethanwharris: Not trainable for now :p Of course, you could always override the training step here to implement some custom logic for training with embeddings, but it shouldn't run without having any loss or optimizer etc. so needs some kind of error here.

tests/core/test_model.py
ethanwharris: > Or at least remove the initial attribute name within the task.

Definitely, I think there are a few enhancements here that would make it easier for users to know which layer to choose for embedding. It would also be cute if we could add a `backbone` layer or something that just goes through everything except the model head, but we may get that for free due to our naming conventions.

I'll leave these enhancements out of this PR as I need to play around with it a bit to figure out what the challenges are, but definitely some things we should look in to!

tests/core/test_model.py
ethanwharris: Good call! Done :smiley:

PR #1421, opened 2022-08-18
Convert `DataKeys` enums to strings in the collate function

### LINKED ISSUE #1343: Revisit TPU training
## 🚀 Feature

just curious, have you used TPU on Kaggle recently?
https://www.kaggle.com/code/jirkaborovec/demo-flash-image-classification-on-tpu

### Motivation

Kaggle has offered 20 hours of training which is mostly used with TF of Keras as any other lib is rather difficult, so there is huge potential to take this space by Flash...
Also With Flash, it shall be trivial to start your work with CPU for exploration and then alternate between GPU and TPU offering to finish with the best-trained model :rabbit: 

### Alternatives

Reason: https://pytorch-lightning.readthedocs.io/en/latest/accelerators/tpu_faq.html#unsupported-datatype-transfer-to-tpus

Unsupported data type transfer to TPUs?
```
File "/usr/local/lib/python3.8/dist-packages/torch_xla/utils/utils.py", line 205, in _for_each_instance_rewrite
    v = _for_each_instance_rewrite(result.__dict__[k], select_fn, fn, rwmap)
File "/usr/local/lib/python3.8/dist-packages/torch_xla/utils/utils.py", line 206, in _for_each_instance_rewrite
    result.__dict__[k] = v
TypeError: 'mappingproxy' object does not support item assignment
```

krshrimali:
## What does this PR do?

The motivation was to enable support for TPU training in Flash, and we realized that transferring Enums to XLA from CPU is not supported. A sample repro script:

```python
import torch_xla.core.xla_model as xm

# Using a plain Enum to show that it's not something related to LightningEnum
class Check(Enum):
    INPUT = 1
    OUTPUT = 2

device = torch.device("xla")
batch = {Check.INPUT: torch.tensor(1)}
print(xm.send_cpu_data_to_device(batch, device)) # This errors out
```

After discussing internally, with @ethanwharris and @Borda - an appropriate solution for this would be to not pass Enums, and convert to strings in the collate function itself. This PR addresses the same.

Partially fixes #1343

## Before submitting

## PR review

## Did you have fun?

cc: @ethanwharris @Borda @kaushikb11


ethanwharris (commented):
Awesome work @krshrimali a <sub><sup>small</sup></sub> fix for a **big** bug! Maybe better to move the enum handling to the `Input` class here: https://github.com/Lightning-AI/lightning-flash/blob/0e2125926d10207ce7e1f905d4c392a465558651/flash/core/data/io/input.py#L174

That way even if you just get the dataset and iterate over it you won't see the enums :smiley:

flash/core/data/utilities/collate.py
```diff
@@ -26,7 +26,8 @@ def _wrap_collate(collate: Callable, batch: List[Any]) -> Any:
     collated_batch = collate(batch)
 
     if metadata and isinstance(collated_batch, dict):
-        collated_batch[DataKeys.METADATA] = metadata
+        collated_batch["metadata"] = metadata
```
ethanwharris, line 29: The `DataKeys.METADATA` should still work here I think (the enum is meant to be interchangeable with a plain string)

krshrimali: You are right. One question, @ethanwharris - once the `datamodule` is created, we don't expect to see the enums after this PR, right? The reason I ask this is - because I can see the Enums even after this change in the PR, for the `datamodule.train_dataloader().dataset`.

ethanwharris: Yeah, after the change you just made I would expect the dataset to not have the enums. Note that this may only be after the most recent change. Before that I think the default `train_dataloader` implementation doesn't use our collate function but not entirely sure.

krshrimali: Thank you for the clarification! Looks like once the `datamodule` is created, the following:

```python
[x.keys() for x in datamodule.train_dataloader().dataset]
```

Gives: a list of: `dict_keys([<DataKeys.INPUT: 'input'>, <DataKeys.TARGET: 'target'>, <DataKeys.METADATA: 'metadata'>])`, which is not what we want/expected, right? Note that this is from this branch, and from the most recent commit.

The functions we made the changes in, only get called once we call the `fit` function from the `Trainer`, as I think then this `_load_sample` method is called from the `InputBase`.

I'll work on replacing enums with strings to ensure that `datamodule` doesn't contain Enums even before the `fit` function is called. I'll update this PR tomorrow! :)

ethanwharris: The `Input` is just a torch dataset. So the `datamodule.train_dataloader().dataset` should be an instance of `InputBase`. The `load_sample` is called in the `getitem` here:
https://github.com/Lightning-AI/lightning-flash/blob/0e2125926d10207ce7e1f905d4c392a465558651/flash/core/data/io/input.py#L277

So in theory if the enums are removed there then they shouldn't turn up in the dataset, but it's possible I'm missing something.

krshrimali: Thanks, @ethanwharris. After investigation, `to_samples` method needed changes, but that only changes `input, target` keys. For `metadata`, every time `_call_load_sample` is called, a `<running_stage>_load_sample` method is called which eventually calls `load_sample` method (https://github.com/Lightning-AI/lightning-flash/blob/0e2125926d10207ce7e1f905d4c392a465558651/flash/core/data/io/input.py#L240).

Changes had to be made to the `load_sample` methods for each task, as they are overridden for each task. I'm thinking to add a test to make sure that this isn't missed out on any task. Will do this tomorrow, but hopefully the changes look good so far?

flash/core/data/io/input.py
```diff
@@ -171,7 +172,19 @@ def __init__(self, running_stage: RunningStage, *args: Any, **kwargs: Any) -> No
...
+        output_dict = {}
+        if isinstance(sample_output, dict):
+            for key, val in sample_output.items():
+                if isinstance(key, Enum) and hasattr(key, "value"):
+                    output_dict[key.value] = val
+                else:
+                    output_dict[key] = val
+        else:
+            return sample_output
+        return output_dict
```
ethanwharris, line 187:
```suggestion
        # Change DataKeys Enum to strings
        if isinstance(sample_output, dict):
            output_dict = {}
            for key, val in sample_output.items():
                if isinstance(key, Enum) and hasattr(key, "value"):
                    output_dict[key.value] = val
                else:
                    output_dict[key] = val
            return output_dict
        return sample_output
```

ethanwharris (approved):
LGTM, small suggestion 😃

PR #1464, opened 2022-10-02
refactored image_classifier_head to classifier_head

### LINKED ISSUE #1462: Refactor `IMAGE_CLASSIFIER_HEADS` to be general classification heads
## 🚀 Feature

We have the `IMAGE_CLASSIFIER_HEADS` registry in `image/classification/heads.py` but it's actually useful for all classification tasks where you have an embedding followed by a classification layer. We should refactor it out and rename it to just `CLASSIFIER_HEADS` in `core/classification.py`

Abelarm:
## What does this PR do?

Refactored the file `image/classification/heads.py` moving the head to the `core/classification.py` changing the name from `IMAGE_CLASSIFIER_HEADS` to `CLASSIFIER_HEADS`

Fixes #1462 

## Before submitting

## PR review

## Did you have fun?
Al lot, hope to "get to know more" the code


ligaz:
Have you considered leaving out the `IMAGE_CLASSIFIER_HEADS` and just instantiate it with all heads from `CLASSIFIER_HEADS`? 

This way you will gain the following:
* Not introduce a breaking change - removing `IMAGE_CLASSIFIER_HEADS`
* Have `IMAGE_CLASSIFIER_HEADS` that could be further extended in the future with image specific heads.

Abelarm:
> Have you considered leaving out the `IMAGE_CLASSIFIER_HEADS` and just instantiate it with all heads from `CLASSIFIER_HEADS`?
> 
> This way you will gain the following:
> 
>     * Not introduce a breaking change - removing `IMAGE_CLASSIFIER_HEADS`
> 
>     * Have `IMAGE_CLASSIFIER_HEADS` that could be further extended in the future with image specific heads.

Regarding the error, pretty sure I didn't touch anything on this error looks like more of a precision error.
```
=================================== FAILURES ===================================
________________________ test_embedder_raising_overhead ________________________

    @pytest.mark.skipif(not _CORE_TESTING, reason="Not testing core.")
    def test_embedder_raising_overhead():
        """Tests that embedding to the output layer of a 3 layer model takes less than 10ms more than the time taken to
        execute the model without the embedder.
    
        Note that this bound is intentionally high in an effort to reduce the flakiness of the test.
        """
        model = NLayerModel(10)
        embedder = Embedder(model, "output")
    
        start = time.perf_counter()
        model.predict_step(torch.rand(10, 1000), 0, 0)
        end = time.perf_counter()
    
        model_time = end - start
    
        start = time.perf_counter()
        embedder.predict_step(torch.rand(10, 1000), 0, 0)
        end = time.perf_counter()
    
        embedder_time = end - start
    
>       assert abs(embedder_time - model_time) < 0.01
E       assert 0.013533204999987447 < 0.01
E        +  where 0.013533204999987447 = abs((0.11517994799999087 - 0.12871315299997832))

tests/core/utilities/test_embedder.py:108: AssertionError
```

Regarding the second point:

I am not an expert of `lightning-flash` I just went with was stated in the Issue #1464, but I don't get your point? 
there were no  `CLASSIFIER_HEADS`

ligaz:
My point is that we are changing the public API by removing the `IMAGE_CLASSIFIER_HEADS`. This means that if there were users that was using that their code will be broken. My suggestion is to not break the public API but rather make it backwards compatible.

@ethanwharris What's your take on this as a original author of issue #1462?

krshrimali:
> My point is that we are changing the public API by removing the `IMAGE_CLASSIFIER_HEADS`. This means that if there were users that were using this API that their code will be broken. My suggestion is to not break the public API but rather make it backwards compatible.
> 
> @ethanwharris What's your take on this as a original author of issue #1462?

I agree, and let's safely assume that there will be users using `IMAGE_CLASSIFIER_HEADS` (though I didn't see relevant usage on GitHub global grep, but that's not enough to say no one uses it).

Let's point `IMAGE_CLASSIFIER_HEADS` to `CLASSIFIER_HEADS` in this PR, and add a deprecation message, something like:

```
In future releases, IMAGE_CLASSIFIER_HEADS will be deprecated in favor of CLASSIFIER_HEADS. Please consider using `<imports>.CLASSIFIER_HEADS` instead.
```

Sorry for the delay in getting back on this. @ethanwharris - Does this sound good to you? Also @ligaz @Abelarm please feel free to share any thoughts. :)

Abelarm:
I am not an expert so I trust your opinion more 🙃

ligaz:
I'm leaning more towards:
> The other alternate way would be to keep IMAGE_CLASSIFIER_HEADS and not deprecate, if in the future there is a possibility that IMAGE_CLASSIFIER_HEADS and CLASSIFIER_HEADS will diverge (or one of them will be a superset of the other).

I'm also perfectly fine with safely deprecating `IMAGE_CLASSIFIER_HEADS`.

Abelarm:
Hi @ligaz and @krshrimali would something like this be an acceptable solution?
My idea was:
- move the "generic_head" to `CLASSIFIER_HEADS`
- keep the `IMAGE_CLASSIFIER_HEADS` but make it point to the generic `CLASSIFIER_HEADS` 

It should keep the backwards compatibility and should be possible to add "image-specific" heads

krshrimali:
Hey, @Abelarm - Yes! I think that's something we all can have a consensus on. :) Looks like you've already made the required changes, I'm sorry for the delay - I'll take this from here. Just need to check if we are missing anything in the docs etc. Thank you for your contribution 🚀 ⚡ 

@ligaz @ethanwharris - Please let me know if you've any further thoughts on this. Thank you all for the active discussion and involvement.

flash/image/classification/model.py
```diff
@@ -32,7 +32,7 @@
 )
 from flash.image.classification.adapters import TRAINING_STRATEGIES
 from flash.image.classification.backbones import IMAGE_CLASSIFIER_BACKBONES
-from flash.image.classification.heads import IMAGE_CLASSIFIER_HEADS
+from flash.image.classification.head import IMAGE_CLASSIFIER_HEADS
```
krshrimali, line 35: I think this was a typo? should be `heads`, I guess:

```suggestion
from flash.image.classification.heads import IMAGE_CLASSIFIER_HEADS
```

flash/text/classification/adapters.py
```diff
@@ -21,11 +21,11 @@
 from torch import Tensor
 
 from flash.core.adapter import Adapter, AdapterTask
+from flash.core.classification import CLASSIFIER_HEADS
```
krshrimali, line 24: Nice!

krshrimali (approved):
LGTM for the changes so far, I'll take another look if we need to make changes anywhere else tonight, and merge if everything looks good. Thanks @Abelarm 🚀

flash/image/classification/model.py
Abelarm: OK this is solved with the latest commit, the name of the file was `head` and not `heads` so the typo was inside the file name not the import

flash/text/classification/adapters.py
Abelarm: This as is like this because I added the `CLASSIFIER_HEADS` inside the `core/classification.py` should I move it to its own file `core/heads.py`

krshrimali: Yeah, actually let's do it. Please move it to its own file: `core/heads.py`. Just to be consistent with other imports. :)

Abelarm:
Looking at the azure pipeline I found out that there is an error:

```
File "/__w/1/s/flash/core/registry.py", line 123, in _register_function
    raise MisconfigurationException(f"You can only register a callable, found: {fn}")
pytorch_lightning.utilities.exceptions.MisconfigurationException: You can only register a callable, found: tensor
```

Could it be that I introduce it?

flash/image/classification/heads.py
```diff
@@ -11,30 +11,12 @@
...
-    """
-    return nn.Linear(num_features, num_classes)
-
-
-IMAGE_CLASSIFIER_HEADS(
-    partial(_load_linear_head),
-    name="linear",
-)
+# Add classifier heads to IMAGE_CLASSIFIER_HEADS using map
+any(map(lambda x: IMAGE_CLASSIFIER_HEADS(x["fn"], name=x["name"]), CLASSIFIER_HEADS.functions))
```
ethanwharris, line 22: A simpler thing you could do here is:
```suggestion
IMAGE_CLASSIFIER_HEADS = IMAGE_CLASSIFIER_HEADS + CLASSIFIER_HEADS
```

ethanwharris (commented):
Thanks for this! Minor suggestion :smiley:

flash/image/classification/heads.py
Abelarm: I didn't really know how it was working the registry so I create a plain new one (from the old) but if I can do this... 🤷 

"The easiest solution it's usually the right one"

flash/image/classification/heads.py
```diff
@@ -11,30 +11,12 @@
...
-        nn.Module: Linear head.
-    """
-    return nn.Linear(num_features, num_classes)
-
-
-IMAGE_CLASSIFIER_HEADS(
-    partial(_load_linear_head),
-    name="linear",
-)
+# Add classifier heads to IMAGE_CLASSIFIER_HEADS using map
```
ethanwharris, line 21:
```suggestion
```

ethanwharris (approved):

PR #1476, opened 2022-10-27
CI: switch GPU pool

Borda:
## What does this PR do?

let's try to switch to Lit pool, and run only multi-GPU setting
in #1181 I have added parametrization for 1 or 2 GPUs but don't remember why... :otter: 

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/.azure-pipelines/testing-template.yml
+++ b/.azure-pipelines/testing-template.yml
@@ -26,7 +26,7 @@ jobs:
       - bash: |
           gpu_inds=$(DEVICES)
           echo $gpu_inds
-          gpus=$( python -c 'print(",".join("$gpu_inds".split(",")[:int(${{ngpus}})]))' )
+          gpus=$( python -c 'print(",".join("${gpu_inds}".split(",")[:${{ngpus}}]))' )
           echo $gpus
           echo "##vso[task.setvariable variable=CUDA_VISIBLE_DEVICES]$gpus"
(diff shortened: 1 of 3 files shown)


.azure-pipelines/testing-template.yml
```diff
@@ -1,30 +1,35 @@
 jobs:
-- ${{ each gids in parameters.gpu_inds }}:
-  - ${{ each config in parameters.configs }}:
+  - ${{ each dom in parameters.domains }}:
     - job:
-      displayName: "domain ${{config}} with GPUs ${{gids}}"
+      displayName: "domain ${{dom}} with 2 GPU"
```
krshrimali, line 4: So we'll always have 2 GPUs? Or is there a chance that we can have 4? :D Maybe we can avoid hard-coding here? Just curious. :) You know the best :)

krshrimali (approved):

.azure-pipelines/testing-template.yml
Borda: The machine even before, always had 2 GPU :thinking:

justusschock: We don't want 4 here. We only want two to be able to test multi-gpu. Having more gpus would reduce the number of parallel workers which would reduce the throughput as these gpus are not under heavy load and thus increase the queues

PR #1507, opened 2023-01-06
Numpy

souravraha:
## What does this PR do?

Handles the loading of `.npy` images of integer data-types better. Previously, the ndarrays would be unsafely cast into `'uint8'`. If the original dtype were, say `'int64'`, the loaded PIL Image would be different from what was intended.

Now it should throw up a warning whenever there is unsafe casting. In addition, it would attempt to load a _float_ ndarray (instead of possibly an _integer_ array).

## Before submitting

## PR review

## Did you have fun?


train_annotations.json
```diff
@@ -0,0 +1 @@
+{"annotations": [{"area": 50, "bbox": [10, 20, 5, 10], "num_keypoints": 2, "keypoints": [10, 15, 2, 20, 30, 2], "category_id": 1, "id": 1, "image_id": 1, "iscrowd": 0}, {"area": 100, "bbox": [20, 30, 10, 10], "num_keypoints": 2, "keypoints": [20, 30, 2, 30, 40, 2], "category_id": 2, "id": 2, "image_id": 2, "iscrowd": 0}, {"area": 125, "bbox": [10, 20, 5, 25], "num_keypoints": 2, "keypoints": [10, 15, 2, 20, 45, 2], "category_id": 1, "id": 3, "image_id": 3, "iscrowd": 0}], "categories": [{"id": 1, "name": "cat", "supercategory": "cat", "keypoints": ["left ear", "right ear"]}, {"id": 2, "name": "dog", "supercategory": "dog", "keypoints": ["left ear", "right ear"]}], "images": [{"file_name": "image_1.png", "height": 64, "width": 64, "id": 1}, {"file_name": "image_2.png", "height": 64, "width": 64, "id": 2}, {"file_name": "image_3.png", "height": 64, "width": 64, "id": 3}]}
```
Borda, line 1: where is this used?

souravraha: I am not sure. This was generated after I ran tests locally. I believe there were a few more. Should be safe to delete, I think.

souravraha: I deleted the unwanted files.

souravraha:
The old contribution is discarded, as it was failing a required test.

Now, the pixel values are scaled linearly to lie in between 0 and 255. This ensures minimal losses while loading an _ndarray_ of dtype **_int64_** to a PIL image. The downside for this is that every _ndarray_ would get scaled  differently, i.e. this is a sample specific scaling (rather than, say a global scaling for samples of the entire dataset).

flash/core/data/utilities/loading.py
```diff
@@ -73,7 +73,13 @@ def _load_image_from_image(file):
 
 
 def _load_image_from_numpy(file):
-    return Image.fromarray(np.load(file).astype("uint8")).convert("RGB")
+    arr = np.load(file)
+    if not (arr == arr.astype("uint8")).all():
```
Borda, line 77: This is not very controlled/consistent, why only uint8 shall be scaled to 0-255 and not float from 0-1?

Borda (changes requested):

